PR #2, opened 2018-08-31
Formatting and grammar fixes for README.md

ErichDonGubler:
You haven't actually asked for this, so don't feel bad if there's something in here you'd like to change or ignore altogether! :)

I've tried to keep the commits as focused as possible, which should make it easy to peruse them and figure out what's actually interesting to you. All in all, I'm really excited by this library -- otherwise, I wouldn't have put time into making this PR!

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,29 +3,29 @@
 [![Crates.io](https://img.shields.io/crates/v/roxmltree.svg)](https://crates.io/crates/roxmltree)
 [![Documentation](https://docs.rs/roxmltree/badge.svg)](https://docs.rs/roxmltree)
 
-Represent an [XML 1.0](https://www.w3.org/TR/xml/) document as a read-only tree.
+Represents an [XML 1.0](https://www.w3.org/TR/xml/) document as a read-only tree.
 
 ## Why read-only?
 
-Because in some cases all you need is to retrieve some data from the XML document.
+Because in some cases all you need is to retrieve some data from an XML document.
 And for such cases, we can make a lot of optimizations.
 
 As for *roxmltree*, it's fast not only because it's read-only, but also because
-it uses [xmlparser], which is times faster then [xml-rs].
-See [Performance](#performance) section for details.
+it uses [xmlparser], which is many times faster than [xml-rs].
+See the [Performance](#performance) section for details.
 
 ## Parsing behavior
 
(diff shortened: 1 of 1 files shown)


RazrFalcon:
Thanks. There are a lot of changes. I'm no sure I can discard a specific commit (at least via github itself), but I want to keep the table formatting as is and keep crate names as italic, and not pre/code (just a personal preference).

So I'm interested only in 5fa1467 commit. Is it ok for you?

ErichDonGubler:
Re: crate names: I can replace backticks in `0f76a9f ` (first commit) with asterisks, if you want -- it was the inconsistency that bothered me rather than things having to be using `pre` formatting. If you're still not interested in changing that, no worries!

I can `git rebase -i` or `cherry-pick` in order to get only the commits you want, so no worries. Git doesn't scare me. :)

RazrFalcon:
>it was the inconsistency

Inconsistency with a usual formatting?

ErichDonGubler:
> Inconsistency with a usual formatting?

I'm actually referring to the `README.md` here -- perhaps you have a system that I don't understand and it looks inconsistent to me. The rule I understand is this: **Crate names should be italic.** But there are a few places this isn't true (emphasis mine):

* Line 14 and 67: `it uses `**`[xmlparser]`**`, which is times faster then `**`[xml-rs]`**`.` (no italics here, but these are definitely crates)
* Line 21: `Python's `**`[lxml](https://lxml.de/)`**` behavior.` (Maybe you prefer something else for a Python project, since it's not in the Rust ecosystem?)
* Line 24, 66, and 130: `Fo more details see `**`[docs/parsing.md](https://github.com/RazrFalcon/roxmltree/blob/master/docs/parsing.md).`** (This is a source file and not a crate, so maybe you don't prefer that styling here.)

RazrFalcon:
The rule is that a crate name should be distinct. That's the main point.

Could you keep only the grammar fixes and merge everything into a single commit? I will think about formatting later.

Thanks for you work anyway. I glad you find this library useful.

ErichDonGubler:
Alright, I fixed this PR to only have the grammar fixes. Let me know if there's anything else I can do for your docs -- if you want a grammar pass on your API docs I can do that too! :)

RazrFalcon:
Thanks again.

Yes. Any grammar fixes are welcome.

PR #8, opened 2018-12-21
Use longer lifetime when returning name and namespace

tmiasko:


RazrFalcon:
Interesting. Didn't thought about it. 

Shouldn't it be done for all `Node` methods?

tmiasko:
Indeed, there are a few more to fix. I will add a more generic test to catch them,
since sometimes it is non-obvious from just reading the source code.

tmiasko:
I changed some more. In the case I hit another one, I will get back to you :-).

RazrFalcon:
Thanks a lot again!

PR #16, opened 2019-11-07
Resolve lifetime error in `Document::root_element`

myrrlyn:
There are three commits in this patch; only the first is material. The other two may be removed or kept at your discretion.

Currently, the following code fails to compile:

```rust
use roxmltree::{Document, Node};

pub struct Manifest<'src> {
  xml: Document<'src>,
}

impl<'src> Manifest<'src> {
  pub fn root_node(&self) -> Node<'_, 'src> {
    self.xml.root_element()
  }
}
```

because of lifetime mismatches. However, the *equivalent* code,

```rust
impl<'src> Manifest<'src> {
  pub fn root_node(&self) -> Node<'_, 'src> {
    self.xml.root().first_element_child().unwrap()
  }
}
```

*does* compile.

This is because the definitions of `Document::root` and `Node::first_element_child` explicitly thread lifetimes, but the definition of `Document::root_element` does not. This causes the compiler to incorrectly attach shorter lifetimes to the returned `Node` of `Document::root_element`.

This patch explicitly provides `'input` and `'a` lifetimes to `Document::root_element`, matching `Document::root`, so that this error is resolved.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -100,7 +100,7 @@ impl<'input> Document<'input> {
     /// assert!(doc.root_element().has_tag_name("e"));
     /// ```
     #[inline]
-    pub fn root_element(&self) -> Node {
+    pub fn root_element<'a>(&'a self) -> Node<'a, 'input> {
         // `unwrap` is safe, because the `Document` is guarantee to have at least one element.
         self.root().first_element_child().unwrap()
     }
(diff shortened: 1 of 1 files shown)


RazrFalcon:
Thanks. I think you should discard the last commit, since clippy/plugin attributes doesn't work on an older Rust. And you should change `unwrap` to `expect` in the comment above the line you edited.

myrrlyn:
Fair enough, and good catch, respectively

RazrFalcon:
Thanks again. I will publish a new version soon.

PR #17, opened 2019-11-13
Use unconstrained lifetimes for the `attribute` functions

myrrlyn:
This commit adds unconstrained input lifetimes to the name parameters
passed in to the attribute lookup functions, so that the search text
does not need to be coupled to the lifetime of the XML parsed structures
being searched. Because the search text is only used for the duration of
the function call, and does not escape in the returned value, the search
term does not need to match the parsed lifetime.


RazrFalcon:
Interesting. Thanks!

myrrlyn:
Thanks for the crate, by the way! I'm writing an XML processor and this is making it incredibly easy to pull out the parts of the document I want with minimal effort on my part.

RazrFalcon:
Glad to hear.

PR #21, opened 2020-03-17
Expose `text` in Document as getter `raw_text()`

SirJosh3917:
I don't know if the benefits outweigh the downsides of allowing callers to use the input string and coupling yourself to a specific type of input (if you wanted to be able to parse XML documents from the network for example in the future or something), but for me, this would help out.

My personal use case is to be able to get at the raw `&str` that makes up a given `Node`, without having to know the document (so `&node.document().raw_text()[node.range()]`). I can then pass around the `&str` without worrying about the borrow checker, and reparse the node when I want to consume the XML data.

(Note: I'm not sure if that's the best way to go about doing things, I'm just a beginner rustacean.)


src/lib.rs
```diff
@@ -134,6 +134,21 @@ impl<'input> Document<'input> {
     pub fn text_pos_at(&self, pos: usize) -> TextPos {
         xmlparser::Stream::from(self.text).gen_text_pos_from(pos)
     }
+
+    /// Returns the raw text of the original document.
```
RazrFalcon: I would prefer *input* rather than *raw*. So it will look like: 

Returns the input text that was used to parse the document.

SirJosh3917: Should the name of `raw_text` be changed to `input_text` as well?

RazrFalcon: Yes.

RazrFalcon:
Test is unnecessary.

RazrFalcon:
Almost done. Can you revert the removed new line?

RazrFalcon:
You forgot to update the test.

SirJosh3917:
?

RazrFalcon:
Nice! Thanks for contributing.

SirJosh3917:
wait i forgot ending code ticks for the documentation too

RazrFalcon:
It doesn't affect documentation generation, so don't worry. I will fix it.

PR #23, opened 2020-04-01
Change Rc to Arc

tomjw64:
Making this change should allow multi-threading with libraries like rayon.
For example, the following code compiles in this branch but not on `master`:
```rust
use std::fs;
use roxmltree::Document;
use rayon::prelude::*;

fn main() {
  let xml = fs::read_to_string("./document.xhtml").unwrap();
  let doc = Document::parse(&xml).unwrap();
  let root = doc.root();

  (0..1000).into_par_iter().for_each(|_| {
    root.descendants().filter(|node| {
      node.tag_name().name() == "p"
    }).count();
  });
}
```
The speed up from `into_par_iter` can be very significant (3.7s vs 14.2s on my machine) if you are walking the tree a bunch of times (let's pretend that we are looking for different items each iteration in the example above).

After running `cargo bench` and running the non-parallel version of this code through both this branch and `master`, there was no significant difference in the run times. So this seems like it might a lot a benefit in some use cases for very little cost.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -23,7 +23,7 @@ extern crate xmlparser;
 use std::borrow::Cow;
 use std::fmt;
 use std::ops::Deref;
-use std::rc::Rc;
+use std::sync::Arc;
 
 pub use xmlparser::TextPos;
 
@@ -469,12 +469,12 @@ impl<'input> Deref for Namespaces<'input> {
 }
 
 
-struct Uri(Rc<String>);
+struct Uri(Arc<String>);
 
 impl Uri {
     #[inline]
     fn new(text: String) -> Self {
-        Uri(Rc::new(text))
+        Uri(Arc::new(text))
     }
 
     #[inline]
@@ -486,7 +486,7 @@ impl Uri {
 impl Clone for Uri {
     #[inline]
     fn clone(&self) -> Self {
-        Uri(Rc::clone(&self.0))
+        Uri(Arc::clone(&self.0))
     }
 }
 


RazrFalcon:
I guess there is nothing that stops us, but could you show an error with `Rc`? I'm interested in what exactly it doesn't like about `Rc`.

tomjw64:
Here is a snippet of the error:
```
error[E0277]: `std::rc::Rc<std::string::String>` cannot be shared between threads safely
  --> src/main.rs:10:29
   |
10 |     (0..1000).into_par_iter().for_each(|_| {
   |  _____________________________^^^^^^^^_-
   | |                             |
   | |                             `std::rc::Rc<std::string::String>` cannot be shared between threads safely
11 | |     root.descendants().filter(|node| {
12 | |       node.tag_name().name() == "p"
13 | |     }).count();
14 | |   });
   | |___- within this `[closure@src/main.rs:10:38: 14:4 root:&roxmltree::Node<'_, '_>]`
   |
   = help: within `[closure@src/main.rs:10:38: 14:4 root:&roxmltree::Node<'_, '_>]`, the trait `std::marker::Sync` is not implemented for `std::rc::Rc<std::string::String>`
   = note: required because it appears within the type `roxmltree::Uri`
```

RazrFalcon:
I've ran the bechmarks and it's 5-10% slower, which is pretty serious. I will think if there is another way, since we don't have a mutable access there is no need in `Arc`.

tomjw64:
Curious. I must be doing something wrong while I'm benchmarking. Here were my results for completeness sake.
Before (w/Rc):
```
test large_roxmltree     ... bench:   4,990,875 ns/iter (+/- 372,508)
test medium_roxmltree    ... bench:     980,338 ns/iter (+/- 25,299)
test tiny_roxmltree      ... bench:       4,890 ns/iter (+/- 237)
```
After (w/Arc):
```
test large_roxmltree     ... bench:   4,753,714 ns/iter (+/- 149,681)
test medium_roxmltree    ... bench:     922,787 ns/iter (+/- 216,640)
test tiny_roxmltree      ... bench:       4,944 ns/iter (+/- 305)
```
I took it to mean that there was no difference, since everything was within the deviation. Any tips for making my benchmarking more reliable?

Good point that there should be no use for `Arc`. Let me see if I can whip anything up as well.

RazrFalcon:
Yes, the measurements are a bit unstable. After multiple runs it looks like it can even be faster in some cases. 

The problem is that benches are testing only the parsing, but `Arc` will affect the public API itself, since every access should be atomic now.

I guess I will simply replace `Uri` with `Cow`. It makes `large_roxmltree` a bit slower, for whatever reason, but `medium_roxmltree` and `tiny_roxmltree` are faster than Rc/Arc now. So it's a fair tradeoff.

The only downside is that when an XML have a lot of not well-formed namespaces (e.g. it has `&\t\n\r`) than it will use way more memory. The idea behind `Uri` is to reduce the memory overhead of duplicated namespaces. But as long as we have a well-formed ns, we simply storing a `&str`.

RazrFalcon:
I've pushed a new commit. Please try it out.

tomjw64:
It works great! Hopefully you don't mind if I send in another PR here and there, even though I am probably less experienced. I have really enjoyed using this library and hope to give back! Thank you!

tomjw64:
Closing in favor of https://github.com/RazrFalcon/roxmltree/commit/13fa4789584ada2f26cf37837a5b26e37f63cec1

tomjw64:
@RazrFalcon I thought you might want to know, that with the change you pushed, I tried `cargo clippy` and it suggested the following:
```
warning: calling `to_string` on `&&mut std::borrow::Cow<'input, str>`
   --> src/parse.rs:798:49
    |
798 |                         *prev_text = Cow::Owned(prev_text.to_string() + text.borrow());
    |                                                 ^^^^^^^^^^^^^^^^^^^^^ help: try dereferencing the receiver: `(*prev_text).to_string()`
    |
    = note: `#[warn(clippy::inefficient_to_string)]` on by default
    = help: `&mut std::borrow::Cow<'input, str>` implements `ToString` through a slower blanket impl, but `std::borrow::Cow<'input, str>` has a fast specialization of `ToString`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#inefficient_to_string
```
I'm sure it's not a big issue, but I thought I'd mention, just in case.

RazrFalcon:
>Hopefully you don't mind if I send in another PR here and there

Sure. Any suggestions are always welcome. Just in this case, the simplest solution wasn't the right one.

RazrFalcon:
I've fixed the clippy warnings. Thanks.

PR #24, opened 2020-04-05
Idea: Guarantee NodeId is in document order

tomjw64:
Hi again!

Since a roxmltree document is read-only, it seems like there are some fairly safe ways that we can take advantage of the order of `NodeId`, since their relationships should never change over the lifetime of the tree. I've made this PR as an example to start a discussion of what benefits that might be drawn from guaranteeing that `NodeId` represents the order of the document.

Some benefits:
1. Implementing the `Ord`/`PartialOrd` traits for depth-first order is trivial. If `Hash` is also implemented for Node, then it can potentially be stored in collections like `BTreeMap`, which could be helpful for anyone making complex selections in a document or for anyone writing selection languages that work on a roxmltree document (I know xpath implementation is a non-goal of this project, but this makes it possible for roxmltree act as the backing representation for something that does make selections).
2. We no longer have to store first child in `NodeData` structs, since it should always be at `self.id + 1` if the node has children. This could lead to some memory savings.
3. There are both benefits and drawbacks for replacing `next_sibling` with `next_subtree` (i.e. the `NodeId` which represents the next node in document order that is not a descendant of the current node). The benefit would be that optimization can be made to the `Descendants` iterator to make it much faster, since it simply has to iterate through NodeId values (takes roughly 50% of the time on my system, which seems like colossal savings!). The drawback of this might be that it reverses some of the savings that we get from not storing the first child in `NodeData` because it does about the same amount of work during parsing, and it is less likely to be `None` than `next_sibling` (I don't actually know enough about the layout of `Option` to know if that matters, perhaps the recent change to `NonZeroUSize` for `NodeId` makes it take up the size either way?). The `next_sibling()` function may also take a slight hit in performance. Maybe this sort of optimization is only worth doing for the root node, but I thought I would mention it, since Descendants is typically a fairly expensive operation and I was seeing such large improvements for it.

Thoughts?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -284,7 +284,7 @@ enum NodeKind<'input> {
 struct NodeData<'input> {
     parent: Option<NodeId>,
     prev_sibling: Option<NodeId>,
-    next_sibling: Option<NodeId>,
+    next_subtree: Option<NodeId>,
     last_child: Option<NodeId>,
     kind: NodeKind<'input>,
     range: Range,
@@ -1012,7 +1012,13 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
     /// Returns the next sibling of this node.
     #[inline]
     pub fn next_sibling(&self) -> Option<Self> {
-        self.d.next_sibling.map(|id| self.gen_node(id))
+        self.d.next_subtree
+            .map(|id| self.gen_node(id))
+            .and_then(|node| {
+                let possibly_self = node.d.prev_sibling
+                    .expect("next_subtree will always have a previous sibling");
+                if possibly_self == self.id { Some(node) } else { None }
+            })
     }
 
     /// Returns the next sibling element of this node.
@@ -1045,7 +1051,7 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
     /// Returns true if this node has siblings.
     #[inline]
     pub fn has_siblings(&self) -> bool {
-        self.d.prev_sibling.is_some() || self.d.next_sibling.is_some()
+        self.d.prev_sibling.is_some() || self.next_sibling().is_some()
     }
 
     /// Returns true if this node has children.
@@ -1090,16 +1096,10 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
         Children { front: self.first_child(), back: self.last_child() }
     }
 
-    /// Returns an iterator which traverses the subtree starting at this node.
-    #[inline]
-    pub fn traverse(&self) -> Traverse<'a, 'input> {
-        Traverse { root: *self, edge: None }
-    }
-
     /// Returns an iterator over this node and its descendants.
     #[inline]
     pub fn descendants(&self) -> Descendants<'a, 'input> {
-        Descendants(self.traverse())
+        Descendants::new(*self)
     }
 
     /// Returns node's range in bytes in the original document.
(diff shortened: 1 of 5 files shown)


RazrFalcon:
Hi! This is an interesting direction.

I didn't go very deep into it yet, but why do we need too change `NodeData` and use `awaiting_subtree`? Isn't the current node order the same? Or am I missing something?

>takes roughly 50% of the time on my system, which seems like colossal savings

Would you mind adding a dedicated benchmark for traversal iterators? Only for `roxmltree`, where it traverses `large.plist`.

Also, how does this change affects parsing performance (the current benchmark)?

As for the Node ordering, I don't really like this idea. It's a tree. It doesn't have an order. I think it will only confuse people. For example, I've originally thought that the order is only for direct children of a node, but it's not.

tomjw64:
**NodeData Changes**
> why do we need too change NodeData and use awaiting_subtree? Isn't the current node order the same?

Yes, the current `NodeId` order is already in document order, but that assumption is not being used in any way right now, other than that the root node is always at `NodeId::new(0)`. The changes in `NodeData` and in the parser were to support the optimizations that we can get if we commit to `NodeId`s being in document-order by design. `awaiting_subtree` is used to resolve the `next_subtree` field in NodeData, and you cannot get the speedup in `descendants` without such a field. To summarize, these are two separate things (i.e. I could make a PR for one but not the other): 
- (document-order assumption => we can avoid storing first child in NodeData)
- (document-order assumption + `next_subtree` => `descendants` optimizations)

**Document Ordering**
> As for the Node ordering, I don't really like this idea. It's a tree. It doesn't have an order. I think it will only confuse people. For example, I've originally thought that the order is only for direct children of a node, but it's not.

I understand your aversion. However, there is precedent for such a guarantee in XML and its derivatives. For example, as I'm sure you know, XPath follows such an ordering:
> There is an ordering, document order, defined on all the nodes in the document corresponding to the order in which the first character of the XML representation of each node occurs in the XML representation of the document after expansion of general entities. Thus, the root node will be the first node. Element nodes occur before their children. Thus, document order orders element nodes in order of the occurrence of their start-tag in the XML (after expansion of entities).

DOM selectors do too:
> The querySelectorAll() methods on the Document, DocumentFragment, and Element interfaces must return a NodeList containing all of the matching Element nodes within the subtrees of the context node, **in document order**. If there are no matching nodes, the method must return an empty NodeList. 

CSS evaluation also follows such an ordering as well with counters. The other side of this is that I'm not sure there's any ordering other than document-order that there is much support for in tooling (because of its relative usefulness) - and that's why I feel this would be a fairly uncontroversial way to provide an implementation of `Ord`/`PartialOrd`.

It's up to you to weigh the the benefits and drawbacks of this. It's my belief that supporting a wider set of collections with generally expected behavior (and some other things that might follow) is worth it, but I'm biased, because this is all stuff I've added to my fork for MY use case, which is not everyone's use case :)

**Benchmarking**
Here are the benchmarks on my system for parsing (it might be best for you to try these too, because as we've discovered, we have some differences in benchmark results between our machines - to me, after running the benchmark a ton of times, it seems like there's no significant difference in the parsing times):
_before_
```
test large_roxmltree                           ... bench:   4,580,286 ns/iter (+/- 321,564)
test medium_roxmltree                          ... bench:     849,993 ns/iter (+/- 40,253)
test tiny_roxmltree                            ... bench:       4,469 ns/iter (+/- 264)
```
_after_
```
test large_roxmltree                           ... bench:   4,489,390 ns/iter (+/- 173,220)
test medium_roxmltree                          ... bench:     876,207 ns/iter (+/- 66,404)
test tiny_roxmltree                            ... bench:       4,586 ns/iter (+/- 202)
```
I added some benchmarks to the PR to test the descendant iteration optimizations (on the root node):
_before_
```
test iteration_descendants_expensive_matches   ... bench:     883,169 ns/iter (+/- 84,350)
test iteration_descendants_inexpensive_matches ... bench:     381,663 ns/iter (+/- 27,863)
```
_after_
```
test iteration_descendants_expensive_matches   ... bench:     648,825 ns/iter (+/- 59,014)
test iteration_descendants_inexpensive_matches ... bench:      75,057 ns/iter (+/- 4,888)
```

src/lib.rs
```diff
@@ -984,7 +1027,7 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
     /// Returns the first child of this node.
     #[inline]
     pub fn first_child(&self) -> Option<Self> {
-        self.d.children.map(|(id, _)| self.gen_node(id))
+        self.d.last_child.map(|_| self.gen_node(self.id + 1))
```
RazrFalcon, line 1030: Can't we use just `self.gen_node(self.id + 1)`?

RazrFalcon:
**NodeData Changes**

I agree that we should stick to nodes order and make it mandatory. But I still not quite understand how the new version works. And I have to, to be able to support this library later.

1. How do we know that `id+1` is actually a first child and not a sibling?
1. Can you explain in the simplest terms what `next_subtree` actually stores?

**Document Ordering**

Well, as long as we can point to a spec, it simplifies things a bit. So I guess we can add this feature, but it should be documented first. Just add it to the `Node` struct comment. Something like: this type supports Ord/PartialOrd in a document order as stated in the spec(link to the spec paragraph).

**Benchmarking**

The speedup is definitely worth pursuing. 

For now, everything looks fine. We should settle with the implementation first and then split it into multiple commits. One per each feature. Plus the changelog updates.

src/lib.rs
tomjw64: We can't. It addresses the question you posed: 
> How do we know that id+1 is actually a first child and not a sibling?

We know it's a first child and not a sibling because we check to see if the element has a last child. If it has a last child, it must also have a first child @ `self.id + 1`, if there is no last child, then `self.id + 1` points to the same node as `next_subtree` (which in most cases would be the next sibling)

tomjw64:
I'll start restructuring this into multiple commits and adding documentation this evening.

tomjw64:
@RazrFalcon I've split it into different commits. As for a visual explanation of what `next_subtree` represents, I have made an example tree where the red line points to a node's `next_subtree`:
next_subtree

In terms of the current model, `next_subtree` is:
- the node with NodeId `self.descendants().last().d.id + 1`
- if `self.next_sibling().is_some()`, then `self.next_sibling()`, else if `self.parent().next_sibling().is_some()`, then `self.parent().next_sibling()` else if `self.parent().parent() ...etc.`

It just so happens that we get this value cheaply during parsing. Hopefully this helps explain!

RazrFalcon:
This is way better that I've expected. You definitely put a lot of effort into it. Thanks again!

On the other hand, I feel stupid, but I still do not understand how the `next_subtree` works. I will play with it tomorrow. For now, could you show an XML that produced the graph on your illustration?

tomjw64:
Happy to help!
You may have noticed, my picture is not quite correct, Node 5 and 6 have the wrong numbers (5 should be labeled 6 and 6 should be labeled 5), even though the arrows are correct. Here is XML that represents the image.
```xml
<zero>
  <one>
    <two />
    <three>
      <four>
        <five />
      </four>
      <six />
    </three>
  </one>
  <seven />
</zero>
```

RazrFalcon:
Oh, so it's literary a next subtree :smile: I got it now. Thanks again for explanation.

PR #26, opened 2020-04-11
Impl Hash for Node

tomjw64:
This will allow `Node` structs to be stored in some frequently used collections like `HashMap` and `HashSet`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,6 +24,7 @@ use std::borrow::Cow;
 use std::cmp::Ordering;
 use std::fmt;
 use std::num::NonZeroUsize;
+use std::hash::{Hash, Hasher};
 use std::ops::Deref;
 
 pub use xmlparser::TextPos;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -632,6 +633,14 @@ impl<'a, 'input> Ord for Node<'a, 'input> {
     }
 }
 
+impl<'a, 'input> Hash for Node<'a, 'input> {
```
RazrFalcon, line 636: Can we use anonymous lifetimes here?

```
impl Hash for Node<'_, '_> {
```

tomjw64: Looks like we can! I can change that. Looks like it can be done for Eq, PartialEq, Ord, PartialOrd impls as well. Does it make sense to anonymize those as well? Perhaps in a separate PR?

RazrFalcon: I guess it will be better to do via a separate request.

PR #27, opened 2020-04-11
Add deep_text method to Node

tomjw64:
Added deep_text method and doctest

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -973,6 +973,32 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


RazrFalcon:
I don't think that this is a good idea. Text processing rules in XML are absurd and in most cases this function will return invalid/unexpected results. For example, it doesn't care about `xml:space`.

Another example is SVG, which has it's own whitespaces processing rules.

tomjw64:
Good point. Also if someone (like myself) really needs something this, it is possible for them to add their own function for their own use case, since this only uses the public api. I will close this. Thanks for the feedback!

PR #29, opened 2020-04-16
Expose NodeId

tomjw64:
I expect to get some pushback on this, but currently this is the last thing that I need (for my use case) to use a fork of roxmltree for, instead of using this branch, so I would be upset with myself if I didn't try a PR :)

**Contents**
This PR add three methods:
1. `Document::nth_node` - which takes a usize `n` and returns the Node at `n` in document-order
2. `Node::order_in_document` - which is the reverse operation of `Document::nth_node`, returns the usize `n` for which `Document::nth_node` returns this node
3. `Document::len` - which returns the number of nodes in the document. I don't really care too much about this one, but it seemed related and currently the quickest way to get this value with the public API is `doc.descendants().count()`, which is a bit odd. I used this method for the doctest for `Node::order_in_document` because it was easier :P

**Reasoning**
My particular use case involves needing to serialize Nodes for IPC. I needed a reversible conversion between `Node` and some type `T` in which `T` is serializable and unique per `Node` in a document. Since `NodeId` is stored already, is unique per `Node` in a document, and has an interface that utilizes usize, it seemed like usize could be the type `T` that I am looking for and that public methods could be exposed that both provided me the conversion I need, and have semantic meaning that does not expose internals like `NodeId` itself.

As far as I can tell, this can _technically_ be implemented with the public API of roxmltree, but it would be very inefficient - i.e. it would require doing something like `Vec<Node>` (for usize -> Node) AND a `HashMap<Node, usize>` (for Node -> usize) with potentially every Node in the document. Perhaps I am missing something easier. Any thoughts?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1015,15 +1015,10 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
         }
     }
 
-    #[inline]
-    fn gen_node(&self, id: NodeId) -> Node<'a, 'input> {
-        Node { id, d: &self.doc.nodes[id.get()], doc: self.doc }
-    }
-
     /// Returns the parent of this node.
     #[inline]
     pub fn parent(&self) -> Option<Self> {
-        self.d.parent.map(|id| self.gen_node(id))
+        self.d.parent.map(|id| self.doc.get_node(id).unwrap())
     }
 
     /// Returns the parent element of this node.
@@ -1034,7 +1029,7 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
     /// Returns the previous sibling of this node.
     #[inline]
     pub fn prev_sibling(&self) -> Option<Self> {
-        self.d.prev_sibling.map(|id| self.gen_node(id))
+        self.d.prev_sibling.map(|id| self.doc.get_node(id).unwrap())
     }
 
     /// Returns the previous sibling element of this node.
@@ -1046,7 +1041,7 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
     #[inline]
     pub fn next_sibling(&self) -> Option<Self> {
         self.d.next_subtree
-            .map(|id| self.gen_node(id))
+            .map(|id| self.doc.get_node(id).unwrap())
             .and_then(|node| {
                 let possibly_self = node.d.prev_sibling
                     .expect("next_subtree will always have a previous sibling");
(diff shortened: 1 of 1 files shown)


RazrFalcon:
Isn't it's easier to simply expose the `NodeId`? Then we can have `Document::get_node` (a public version of `gen_node`).

The reason I haven't exposed the `NodeId` is that it's very easy to mess up when you have more than one document.

tomjw64:
Do you suggest keeping the implementation the same, but change the method names and make the caller do the conversion from `usize` to `NodeId`?

It also occurs to me that this code (or code exposing `NodeId`) can panic via integer overflow because the `NodeId::new` unwraps the `NonZeroUsize` it gets. I will have to address that.

RazrFalcon:
I'm proposing to use the same implementation as [ego-tree](https://github.com/causal-agent/ego-tree).

`nth_node` looks like a hack. Do we really need it?

`try_new` is not necessary. It's ok to panic in this case.

`order_in_document` is also unnecessary. `NodeId` already has a required method.

tomjw64:
I pushed a version that exposes `NodeId`. Is this closer to what you were thinking?

RazrFalcon:
Yes, something like this. And we can remove `Node::gen_node` now.

Does this works for you?

src/lib.rs
```diff
@@ -1116,6 +1142,27 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
...
+    /// 
+    /// ```
+    /// let doc = roxmltree::Document::parse("\
+    /// <p>
+    ///     text
+    /// </p>
+    /// ").unwrap();
+    /// 
+    /// use roxmltree::NodeId;
+    /// (0..doc.descendants().count()).for_each(|n| {
```
RazrFalcon, line 1158: This is a very complicated example. We should simplify it or remove it completely.

src/lib.rs
```diff
@@ -253,17 +277,19 @@ pub struct PI<'input> {
 /// Index into a `Tree`-internal `Vec`.
 ///
 /// By using `NonZeroUsize` we can fit `Option<NodeId>` into a single byte.
```
RazrFalcon, line 279: We should remove this line.

src/lib.rs
tomjw64: I will remove it, since this is just a getter-method now, essentially.

tomjw64:
I addressed the comments. I also removed `Node::gen_node` like you said in the second commit.

RazrFalcon:
Thanks!

RazrFalcon:
Hello again. For your use-case, `NodeId::new` and `NodeId::get` should be public? I would like to hide them.

tomjw64:
I need a way to convert between something serializable to Node. NodeId is not serializable (since its field is private), but usize is, so either `NodeId::new`/`NodeId::get` need to be public or some other way of getting a unique, reversible, serializable value per Node would need to exist.

RazrFalcon:
Good. Then I leave it as is.

Issue #25, opened 2020-04-07
Huge memory hog


npajkovsky:
Hello,

I'm trying to use the library for parsing medium size xml, but some of them are gigantic, like this one

   https://vdp.cuzk.cz/vymenny_format/soucasna/20200331_OB_554782_UZSZ.xml.zip

Uzipped xml file is about 982M 20200331_OB_554782_UZSZ.xml.

Now, when you try to run e.g. stats example, it will eat your whole memory, then swap and it will so so finish on my MacBook Pro 2018 16GB ram.

RazrFalcon:
You should use zero allocation parsers, like [xmlparser](https://github.com/RazrFalcon/xmlparser) for large files.

On the other hand, `roxmltree` should usually use less memory that the input data. Look like your file bloats the memory with namespaces.

I will see what can be done.

npajkovsky:
> You should use zero allocation parsers, like [xmlparser](https://github.com/RazrFalcon/xmlparser) for large files.

Thanks for pointing to that. I'll look at it.

> On the other hand, `roxmltree` should usually use less memory that the input data. Look like your file bloats the memory with namespaces.

When I print AST, it basically prints namespaces over and over.

> I will see what can be done.

Thanks.

RazrFalcon:
>When I print AST, it basically prints namespaces over and over.

Yes, right know `roxmltree` resolves namespaces for each node, which is way too expensive for files like this. On the other hand, it makes querying faster. Will see.

npajkovsky:
> > When I print AST, it basically prints namespaces over and over.
> 
> Yes, right know `roxmltree` resolves namespaces for each node, which is way too expensive for files like this. On the other hand, it makes querying faster. Will see.

What do you mean by "query xml"? Can you give me example of simple query? I cannot recall, that I found any word about query in doc :-/

RazrFalcon:
I meant the `Node`'s API.

RazrFalcon:
In the latest master the memory usage have reduced significantly thanks to @tomjw64. It still uses 7.5GiB. which is 7.5 times bigger than the input file, but it should be good for now. 

Feel free to reopen this issue if you think that this is not enought.

PR #30, opened 2020-04-18
Namespace efficiency

tomjw64:
I thought I'd try my hand at addressing #25. This PR probably still needs some work and discussion, but I thought it would serve as a decent proof of concept.

**Rationale**
It seems that documents with either

1. a lot of namespaces
2. a large size with at least a couple namespaces

were being inflated in size in memory.

This PR tries to reduce memory consumption by only storing each _unique_ namespace declaration within `Document.namespaces` storing a `Vec<usize>` for each node instead of a `Range`, containing indices referencing namespaces in `Document.namespaces`. This does not use _strictly_ less memory than before, but provides excellent savings when roxmltree tends to consume lots of memory from namespaces.

**Examples**

Here's a rundown of maximum memory usage on master and on this branch (I used valgrind to measure memory usage):

 _Peak Memory Usage in MB_
| | Master | This Branch |
|----------------------------------------|--------|-------------|
| large.plist (no root namespace)        | 9.8    | 10.1        |
| large.plist (with root namespace)      | 10.6   | 10.3        |
| large.plist (2 root namespaces)        | 11.3   | 10.3        |
| large.plist (3 root namespaces)        | 12.8   | 10.5        |
| | |
| book.xhtml (12 MB + 2 root namespaces) | 217.4  | 135.9       |
| book.xhtml (3 root namespaces)         | 217.4**  | 140.9       |
| book.xhtml (4 root namespaces)         | 313.4  | 140.9       |

** - Not sure why this is the case, but I double checked this one. I have no idea...

**Parsing Performance**
Since we can just merge the namespace indices of the currently parsed node with its parent's, reducing we can get a noticeable speed-up during parsing for large/namespace-rich documents (commit 0c1db61). In particular, the most gains are in `medium` (I assume since it's the only bench document with namespace declarations), with possible benefits for `large`, and potentially a small slowdown in `tiny`

Here are three consecutive benchmark runs:

_BEFORE_
```
test large_roxmltree                        ... bench:   4,500,096 ns/iter (+/- 173,435)
test medium_roxmltree                       ... bench:     948,893 ns/iter (+/- 36,222)
test tiny_roxmltree                         ... bench:       4,640 ns/iter (+/- 131)

test large_roxmltree                        ... bench:   4,825,849 ns/iter (+/- 338,981)
test medium_roxmltree                       ... bench:     991,011 ns/iter (+/- 93,786)
test tiny_roxmltree                         ... bench:       4,906 ns/iter (+/- 244)

test large_roxmltree                        ... bench:   4,469,187 ns/iter (+/- 288,316)
test medium_roxmltree                       ... bench:     938,886 ns/iter (+/- 142,450)
test tiny_roxmltree                         ... bench:       4,515 ns/iter (+/- 212)
```

_AFTER_
```
test large_roxmltree                        ... bench:   4,428,082 ns/iter (+/- 232,850)
test medium_roxmltree                       ... bench:     862,197 ns/iter (+/- 17,036)
test tiny_roxmltree                         ... bench:       4,758 ns/iter (+/- 210)

test large_roxmltree                        ... bench:   4,418,364 ns/iter (+/- 168,545)
test medium_roxmltree                       ... bench:     868,548 ns/iter (+/- 34,701)
test tiny_roxmltree                         ... bench:       4,752 ns/iter (+/- 262)

test large_roxmltree                        ... bench:   4,549,643 ns/iter (+/- 179,861)
test medium_roxmltree                       ... bench:     886,984 ns/iter (+/- 16,533)
test tiny_roxmltree                         ... bench:       4,929 ns/iter (+/- 173)
```

**Drawbacks**
This change is NOT backwards compatible. `Node::namespaces()` can no longer return a `&[Namespace]`, since new continuous allocations are no longer made per each node. Instead a `Vec<&Namespace>` is returned, because the backing `Vec<usize>` is resolved and collected. I'm not sure how to avoid this backward compatibility issue without intentionally leaking memory to provide a slice. I don't know the policy for breaking changes on this project, so there may be some discussion needed.

**Testing**
While this passes all the existing tests, because this seems like a pretty weighty change, it might be a good idea to include a more thorough regression test for this.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -219,7 +219,7 @@ impl<'input> fmt::Debug for Document<'input> {
                     writeln_indented!(depth, f, "Element {{");
                     writeln_indented!(depth, f, "    tag_name: {:?}", child.tag_name());
                     print_vec("attributes", child.attributes(), depth + 1, f)?;
-                    print_vec("namespaces", child.namespaces(), depth + 1, f)?;
+                    print_vec("namespaces", &child.namespaces(), depth + 1, f)?;
 
                     if child.has_children() {
                         writeln_indented!(depth, f, "    children: [");
@@ -298,7 +298,7 @@ enum NodeKind<'input> {
     Element {
         tag_name: ExpandedNameOwned<'input>,
         attributes: Range,
-        namespaces: Range,
+        namespaces: Vec<usize>,
     },
     PI(PI<'input>),
     Comment(&'input str),
@@ -918,12 +918,17 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
     /// assert_eq!(doc.root_element().namespaces().len(), 1);
     /// ```
     #[inline]
-    pub fn namespaces(&self) -> &'a [Namespace<'input>] {
+    pub fn namespaces(&self) -> Vec<&'a Namespace<'input>> {
         match self.d.kind {
             NodeKind::Element { ref namespaces, .. } => {
-                &self.doc.namespaces[namespaces.clone()]
+                namespaces
+                    .iter()
+                    .map(|i| {
+                        &self.doc.namespaces[*i]
+                    })
+                    .collect()
             }
-            _ => &[],
+            _ => vec![],
         }
     }
 
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -571,6 +571,33 @@ fn process_element<'input>(
         }
     }
 
+    match end_token {
+        xmlparser::ElementEnd::Close(prefix, local) => {
+            let prefix = prefix.as_str();
+            let local = local.as_str();
+
+            doc.nodes[parent_id.get()].range.end = token_span.end();
+            if let NodeKind::Element { ref tag_name, .. } = doc.nodes[parent_id.get()].kind {
+                if prefix != tag_name.prefix || local != tag_name.name {
+                    return Err(Error::UnexpectedCloseTag {
+                        expected: gen_qname_string(tag_name.prefix, tag_name.name),
+                        actual: gen_qname_string(prefix, local),
+                        pos: err_pos_from_span(doc.text, token_span),
+                    });
+                }
+            }
+            pd.awaiting_subtree.push(*parent_id);
+
+            if let Some(id) = doc.nodes[parent_id.get()].parent {
+                *parent_id = id;
+            } else {
+                unreachable!("should be already checked by the xmlparser");
+            }
+            return Ok(())
+        }
+        _ => {}
+    };
+
     let namespaces = resolve_namespaces(pd.ns_start_idx, *parent_id, doc);
     pd.ns_start_idx = doc.namespaces.len();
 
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -918,12 +918,17 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
     /// assert_eq!(doc.root_element().namespaces().len(), 1);
     /// ```
     #[inline]
-    pub fn namespaces(&self) -> &'a [Namespace<'input>] {
+    pub fn namespaces(&self) -> Vec<&'a Namespace<'input>> {
```
RazrFalcon, line 921: Sorry, but the public API must not allocate. Only parsing is allowed to allocate. That's the rule.

So this should be rewritten to iterator or something.

RazrFalcon:
Yes, I also thought about it, but I don't see a good solution for now. It obviously will be a breaking changes and this is not a problem.

My idea is to store namespaces as they appear in the document, and resolve them on request. This will dramatically reduce the memory usage, but I bet it will hit the performance too much. So it's not a solution, really. But I'm open for other ideas.

src/lib.rs
```diff
@@ -298,7 +298,7 @@ enum NodeKind<'input> {
     Element {
         tag_name: ExpandedNameOwned<'input>,
         attributes: Range,
-        namespaces: Range,
+        namespaces: Vec<usize>,
```
RazrFalcon, line 301: I'm not sure that this is a good idea. I prefer one large allocation to multiple tiny one. Afaik, the benchmarks can be affected drastically by the used allocator. So it's hard to say that this is actually better before testing on different OS'es with different allocators.

src/lib.rs
tomjw64: Yeah, it was an unfortunate side effect of this PR. It could easily return an iterator instead, as you say, if that is acceptable.

src/lib.rs
tomjw64: Hmmmm... good point that lots of tiny allocations take place here. Is there the potential to limit the number of namespace in scope? I'll keep thinking :thinking:

RazrFalcon: What kind of limit? The maximim number of namespaces per element? I don't that any kind of limit is a good idea. Plus, even 3-5 namespaces in an XML with thousands of elements will blow up the memory.

As always, we have to choose between memory and CPU usage. Right now, `roxmltree` leans to a lower CPU usage, which is basically zero in the current implementation (at least when we work with namespaces).

RazrFalcon:
I guess a better idea is too find a way to somehow compact the current `Vec<Namespace>`. For example, if we have only one namespace on the root element, the current algorithm will still duplicate it for each node. But technically we can store only the first one and set `0..1` ns range for all elements.

Of course, it will not work when document has a lot of mixed namespaces, but it still a good direction.

tomjw64:
I am considering simply reducing duplication in this manner, only cloning namespaces when a node's namespaces != it's parent. So in the worst case (a namespace declaration on every node), it would perform as it does currently. Will open another PR soon.

tomjw64:
Okay #31 I think is a much safer change and does not have the same tradeoffs this branch does.

tomjw64:
Closing in favor of #31

PR #31, opened 2020-04-18
Reduce duplication in Namespaces

tomjw64:
I believe this is strictly better or equal memory efficiency to master and does not require a breaking change.

 _Peak Memory Usage in MB_
| | Master | This Branch |
|----------------------------------------|--------|-------------|
| large.plist (no root namespace)        | 9.8    | 9.8       |
| large.plist (with root namespace)      | 10.6   | 9.8       |
| large.plist (2 root namespaces)        | 11.3   | 9.8       |
| large.plist (3 root namespaces)        | 12.8   | 9.8        |
| | |
| book.xhtml (12 MB + 2 root namespaces) | 217.4  | 121.4       |
| book.xhtml (4 root namespaces)         | 313.4  | 121.4       |

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -645,32 +645,19 @@ fn resolve_namespaces(
     parent_id: NodeId,
     doc: &mut Document,
 ) -> Range {
-    let mut tmp_parent_id = parent_id.get();
-    while tmp_parent_id != 0 {
-        let curr_id = tmp_parent_id;
-        tmp_parent_id = match doc.nodes[tmp_parent_id].parent {
-            Some(id) => id.get(),
-            None => 0,
-        };
-
-        let ns_range = match doc.nodes[curr_id].kind {
-            NodeKind::Element { ref namespaces, .. } => namespaces.clone(),
-            _ => continue,
-        };
-
-        for i in ns_range {
+    if let NodeKind::Element { ref namespaces, .. } = doc.nodes[parent_id.get()].kind {
+        let parent_ns = namespaces.clone();
+        if start_idx == doc.namespaces.len() {
+            return parent_ns
+        }
+        for i in parent_ns {
             if !doc.namespaces.exists(start_idx, doc.namespaces[i].name) {
                 let v = doc.namespaces[i].clone();
                 doc.namespaces.0.push(v);
             }
         }
     }
-
-    if start_idx != doc.namespaces.len() {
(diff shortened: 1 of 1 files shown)


src/parse.rs
```diff
@@ -645,32 +645,22 @@ fn resolve_namespaces(
...
-        tmp_parent_id = match doc.nodes[tmp_parent_id].parent {
-            Some(id) => id.get(),
-            None => 0,
-        };
-
-        let ns_range = match doc.nodes[curr_id].kind {
-            NodeKind::Element { ref namespaces, .. } => namespaces.clone(),
-            _ => continue,
-        };
+    let parent_id = parent_id.get();
```
RazrFalcon, line 648: I guess we can move this line since `parent_id` used only once anyway.

RazrFalcon:
If I understood correctly, we are simply using the parent namespaces as long as the current node didn't added it's own?

tomjw64:
That's correct. So the worst case is namespace declarations being made on every element, in which case the duplication will be the same as it is now.

RazrFalcon:
Looks like this change doesn't break the `resvg`, which is good. And the linked 1GiB XML from #25 peaks at 7.5GiB and not whatever (I've killed it when it reached 25GiB). So this is definitely a successes and an easy one. Thanks again!

RazrFalcon:
Do you have any other ideas or should we publish a new release?

tomjw64:
I don't have anything else right now :)

RazrFalcon:
Nice! Then I will add few more changes and will publish a new release.

tomjw64:
Awesome! Very happy to contribute to this release!

RazrFalcon:
I've added a new commit that reduced memory usage on the linked XML from 7.5GiB to 6.5GiB. It's a bit controversial since it imposes a 4GiB limit and replaces `usize` with `u32`. But if you are fine with this change we can publish a new release right now.

tomjw64:
Hmmm I think it's interesting. Probably pretty safe, since it's doubtful many systems have enough ram to handle anything larger. I'll take a look at the commit.

tomjw64:
I took a look. Seems reasonable to me!

RazrFalcon:
Good! Publishing.

RazrFalcon:
Thanks again for patches.

PR #32, opened 2020-05-15
Make `roxmltree::Error` compare-able

jfrimmel:
Hello,
I'm an active user of this crate. In my current project I need a way to compare the errors returned from the XML tree parsing. To enable this, I created this PR.

This commit additionally derives the equality-related traits. Therefore errors can now be compared. As this is a auto-derive, no tests were added.

__Note__: this is blocked by https://github.com/RazrFalcon/xmlparser/pull/12, as the underlying `xmlparser::Error` has to have the same traits as well. That PR has to be accepted and the dependency version number has to be increased before merging this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ exclude = ["test-apps", "testing-tools"]
 travis-ci = { repository = "RazrFalcon/roxmltree" }
 
 [dependencies]
-xmlparser = "0.13.1"
+xmlparser = "0.13.2"
 
 [dev-dependencies]
 pretty_assertions = "0.5"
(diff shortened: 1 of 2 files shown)


jfrimmel:
Okay, I've used the same style as in the other PR. Before merging this a new version of `xmlparser` has to be released. I can do this dependency upgrade in this PR as well once that is done.

RazrFalcon:
I've completely forgot about it. Since I've just published a new version of xmlparser, you can finish it now.

jfrimmel:
I forgot it myself :smile: 

But it should now be ready. Just say, if I should update the changelog/crate version/etc. as well.

PR #35, opened 2020-06-19
Adjust lifetime of name in ExpandedName

rkusa:
This PR adjusts the lifetime of `ExpandedName::name` to return a string that is bound to the lifetime of the input and not to the lifetime of the ~~`ExpandedName`~~ `ExpandedNameOwned` struct. I've added a test to illustrate an (artificial) use-case that is fixed with this PR:

```rust
    fn get_tag_name<'a, 'input>(node: &'a Node<'a, 'input>) -> &'input str {
        node.tag_name().name()
    }
```

This method would not compile without this PR, it would fail with the following error:

```
error[E0623]: lifetime mismatch
   --> tests/api.rs:290:9
    |
289 |     fn get_tag_name<'a, 'input>(node: &'a Node<'a, 'input>) -> &'input str {
    |                                       --------------------     -----------
    |                                       |
    |                                       this parameter and the return type are declared with different lifetimes...
290 |         node.tag_name().name()
    |         ^^^^^^^^^^^^^^^^^^^^^^ ...but data from `node` is returned here
```

Thanks in advance for considering this change 🙏

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -483,9 +483,9 @@ struct ExpandedNameOwned<'input> {
     name: &'input str,
 }
 
-impl<'input> ExpandedNameOwned<'input> {
+impl<'a, 'input> ExpandedNameOwned<'input> {
     #[inline]
-    fn as_ref(&self) -> ExpandedName {
+    fn as_ref(&'a self) -> ExpandedName<'a, 'input> {
         ExpandedName {
             uri: self.ns.as_ref().map(Cow::as_ref),
             name: self.name,
@@ -507,12 +507,12 @@ impl<'input> fmt::Debug for ExpandedNameOwned<'input> {
 ///
 /// Contains an namespace URI and name pair.
 #[derive(Clone, Copy, PartialEq)]
-pub struct ExpandedName<'input> {
-    uri: Option<&'input str>,
+pub struct ExpandedName<'a, 'input> {
+    uri: Option<&'a str>,
     name: &'input str,
 }
 
-impl<'input> ExpandedName<'input> {
+impl<'a, 'input> ExpandedName<'a, 'input> {
     /// Returns a namespace URI.
     ///
     /// # Examples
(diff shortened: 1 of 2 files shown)


RazrFalcon:
The solution looks overly complicated... But looks like there is no other choice.

src/lib.rs
```diff
@@ -483,9 +483,9 @@ struct ExpandedNameOwned<'input> {
     name: &'input str,
 }
 
-impl<'input> ExpandedNameOwned<'input> {
+impl<'a, 'input> ExpandedNameOwned<'input> {
```
RazrFalcon, line 486: Shouldn't it be `'a, 'input: 'a`?

src/lib.rs
```diff
@@ -507,12 +507,12 @@ impl<'input> fmt::Debug for ExpandedNameOwned<'input> {
 ///
 /// Contains an namespace URI and name pair.
 #[derive(Clone, Copy, PartialEq)]
-pub struct ExpandedName<'input> {
-    uri: Option<&'input str>,
+pub struct ExpandedName<'a, 'input> {
+    uri: Option<&'a str>,
```
RazrFalcon, line 511: Why have you changed the lifetime only for `uri`?

rkusa: Because, as far as I understand it, the lifetime of `name` is the lifetime of the initial input, where the lifetime of `uri` corresponds to the lifetime of `ExpandedNameOwned`, since we borrow an owned value from `ExpandedNameOwned` in line 490 `self.ns.as_ref()`. In other words, because `ExpandedNameOwned` owns `ns` (where `uri` is derived from) and only has  a borrow of `name`. Does this make sense?

src/lib.rs
rkusa: I think it would, from a logical point of view, make sense that `'input` lives at least as long as `'a`, but TBH, I don't know the benefit of adding it if both versions (with and without) compile?

rkusa:
> The solution looks overly complicated... But looks like there is no other choice.

I agree. This is also the reason why I was a bit hesitant to even propose the change (the commit of mine dates back to April). The alternative for me would be to add quite some `.to_string()` to my tool that uses `roxmltree`. But I could totally understand if you don't want this change, and I could live with the additional `.to_string()`s then.

src/lib.rs
RazrFalcon: My bad.

RazrFalcon:
The change makes sense. I'm using roxmltree in a query-only way, so I don't really care about lifetimes. But your proposal is more correct.

Issue #36, opened 2020-06-20
roxmltree v0.11.1 breaks code that worked with v0.11.0


psychon:
Hi,

you broke my build. ;-)

This is just a FYI that something happened. I guess that nothing can be done about it even though v0.11.1 is not semver compatible with v0.11.0 (or I misunderstood something).

The error is:
```
error[E0621]: explicit lifetime required in the type of `name`
    --> xcbgen-rs/src/parser.rs:1129:5
     |
1128 | fn get_attr<'a>(node: roxmltree::Node<'a, '_>, name: &str) -> Result<&'a str, ParseError> {
     |                                                      ---- help: add explicit lifetime `'a` to the type of `name`: `&'a str`
1129 |     node.attribute(name).ok_or(ParseError::InvalidXml)
     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ lifetime `'a` required
error: aborting due to previous error
```
The code in question is this:
https://github.com/psychon/x11rb/blob/7f2407f735881acebb46c93eb3bc8fed3fa99814/xcbgen-rs/src/parser.rs#L1128-L1130
```rust
fn get_attr<'a>(node: roxmltree::Node<'a, '_>, name: &str) -> Result<&'a str, ParseError> {
    node.attribute(name).ok_or(ParseError::InvalidXml)
}
```
I haven't looked closely yet at what happened, but the compiler's suggestion seems wrong to me. Perhaps I need `Node<'a, 'a>`? Dunno, future-me will experiment with this.

Anyway, feel free to just close this issue. As I said, I just wanted to let you know about this.

RazrFalcon:
Yes, I guess 0.11.1 should be yanked and replaced with 0.12. Looks like the last change was actually a breaking one.

As for the fix, the method should look like this:

https://github.com/RazrFalcon/roxmltree/blob/b358cce15ab0663909659f2822ac2c9f6a7d7c9e/tests/api.rs#L269

RazrFalcon:
Done.

psychon:
Thanks for the fix

psychon:
However, it didn't actually help me. Sorry.

`attribute()` returns an `Option<&'a str>`, so adding an `'input` lifetime does not help here. Instead, I have to ensure that the given `&str` lives for `'a`, i.e.:
```diff
[...]

PR #37, opened 2020-06-20
Fix lifetimes when using `ExpandedName` as parameter.

eduardosm:
Now, `ExpandedName<'n, 'm>` can have arbitrary `'n` and `'m` lifetimes.

I also renamed the second lifetime of `ExpandedName` from `'input` to `'b`, as the `name` field might not always point to the XML input data.

Fixes https://github.com/RazrFalcon/roxmltree/issues/36

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -595,12 +595,12 @@ impl<'input> fmt::Debug for ExpandedNameOwned<'input> {
 ///
 /// Contains an namespace URI and name pair.
 #[derive(Clone, Copy, PartialEq)]
-pub struct ExpandedName<'a, 'input> {
+pub struct ExpandedName<'a, 'b> {
     uri: Option<&'a str>,
-    name: &'input str,
+    name: &'b str,
 }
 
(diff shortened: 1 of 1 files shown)


psychon:
Thanks for writing this! :-)

RazrFalcon:
Thanks! Can it be considered a breaking change? Or we can push it as 0.12.1?

psychon:
```diff
-    pub fn has_tag_name<'n, N>(&self, name: N) -> bool
+    pub fn has_tag_name<'n, 'm, N>(&self, name: N) -> bool
```
If I had some code that explicitly did `foo.has_tag_name::<'a, whatever>(name)`, that would no longer compile, right? Because now the number of generic arguments no longer matches?

So, I do not want to make the decision, but at least I "feel like one could construct code" that breaks due to this (but I also doubt that anyone wrote code like this).

RazrFalcon:
Yeah, adding a lifetime is probably a breaking change. 

This "change" already received two updates and making a new one with version *13* doesn't sound like a good idea :smile:

PR #40, opened 2020-12-25
chore: Update pretty_assertions to 0.6

ignatenkobrain:


RazrFalcon:
`pretty_assertions` > 0.6 moved to proc macros on Windows, which takes forever to compile. I'm using old version intentionally.

michel-slm:
Can we use 0.5 on Windows and 0.6 everywhere else?

RazrFalcon:
It would be easier to get rid of `pretty_assertions` altogether.

PR #45, opened 2021-05-20
Fix returned lifetime of `Document::descendants()`.

DBLouis:
Also enable lint `elided_lifetimes_in_paths` (warn).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -142,7 +142,7 @@ impl<'input> Document<'input> {
     ///
     /// Shorthand for `doc.root().descendants()`.
     #[inline]
-    pub fn descendants(&self) -> Descendants {
+    pub fn descendants(&self) -> Descendants<'_, 'input> {
         self.root().descendants()
     }
 


RazrFalcon:
Can you provide a test case for `descendants`?

Also, most of those changes are not necessary. At least I don't see any point in them. Could you explain at least some of them? Like `err_pos_from_span` makes no sense to me.

DBLouis:
> Can you provide a test case for `descendants`?
> 
> Also, most of those changes are not necessary. At least I don't see any point in them. Could you explain at least some of them? Like `err_pos_from_span` makes no sense to me.

Almost all the changes are to satisfy the lint. This fix is only about `Document::descendants()` which had different lifetime than the equivalent function `Node::descendants()`. I just though that enabling this lint could prevent future mistakes of the same kind.

RazrFalcon:
I don't see any reasons to use this lint. It just forces explicit lifetimes. It doesn't actually prevent any bugs, afaik.

>This fix is only about Document::descendants() which had different lifetime than the equivalent function Node::descendants(). 

1. You've also changed `pub fn name(&self) -> &'input str {`
2. Why it's `pub fn descendants(&self) -> Descendants<'_, 'input> {` and not `pub fn descendants<'a>(&'a self) -> Descendants<'a, 'input> {`
3. In what case it causes troubles?

DBLouis:
>     1. You've also changed `pub fn name(&self) -> &'input str {`

I forgot to remove that ;)

>     2. Why it's `pub fn descendants(&self) -> Descendants<'_, 'input> {` and not `pub fn descendants<'a>(&'a self) -> Descendants<'a, 'input> {`

No reason, it's equivalent.

>     3. In what case it causes troubles?

The doc says it's shorthand for `doc.root().descendants()` but it's not strictly the same because the lifetime differs.

RazrFalcon:
Ok. Now it makes sense. Thanks.

PR #51, opened 2021-07-20
Fix a few Clippy warnings

baskerville:


RazrFalcon:
Thanks, but I'm fine with the current implementation. You can update `Rc` comparison and remove some `.clone()` calls, but I would prefer if everything else would stay the same.

src/parse.rs
```diff
@@ -283,10 +283,7 @@ impl<'input> Document<'input> {
     ) -> NodeId {
         let new_child_id = NodeId::from(self.nodes.len());
 
-        let appending_element = match kind {
-            NodeKind::Element {..} => true,
-            _ => false
-        };
+        let appending_element = matches!(kind, NodeKind::Element {..});
```
RazrFalcon, line 286: `matches!` was added in 1.42, while roxmltree supports up to 1.31.

PR #54, opened 2021-10-03
Add function to set allow_dtd in builder style

philippeitis:
Generally, most libraries allow configuring options through functions in the style of `fn value(self, value: ValueType) -> Self`. I have added this to ParsingOptions to make specifying the options more natural - instead of

```rust
let mut options = roxmltree::ParsingOptions::default();
options.allow_dtd = true;
let document = Document::parse_with_options(&document_text, options).unwrap();
```

you can now do
```rust
let document = Document::parse_with_options(
    &document_text,
    roxmltree::ParsingOptions::default()
        .allow_dtd(true)
).unwrap();
```

This builder style would be also useful in the future, if new options are ever added, as they can simply be chained.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -229,6 +229,23 @@ impl Default for ParsingOptions {
     }
 }
 
+impl ParsingOptions {
+    /// Allows DTD parsing if set to `true`. Otherwise, XML with DTD will cause an error.
+    /// Empty DTD block is not an error.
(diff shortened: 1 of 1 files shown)


RazrFalcon:
Sorry, but I don't like the builder pattern in general. That's why I'm not using it. A basic `struct` is way more flexible.

PR #57, opened 2022-07-10
feat: add `namespace` to return uri via name

duskmoon314:
Add a new method to get the URI of a namespace via its name. Pretty much like the `attribute` method.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -982,6 +982,29 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
         }
     }
 
+    /// Returns element's namespace uri
+    ///
+    /// # Examples
+    ///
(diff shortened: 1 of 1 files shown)


RazrFalcon:
Aren't this `lookup_namespace_uri`?

duskmoon314:
I noticed that there was an `attribute` method, and was surprised that there was no `namespace` method. It turns out that I didn't notice that there is a `lookup_namespace_uri`. 😂

My fault. orz

RazrFalcon:
No problem. The method name might sound confusing, but this is an official naming convention: https://developer.mozilla.org/en-US/docs/Web/API/Node/lookupNamespaceURI

Issue #63, opened 2022-10-06
Reduce peak memory usage


RazrFalcon:
Measure and try to reduce the peak memory overhead over the input file. Basically find out how much memory is used by roxmltree metadata itself.

- [x] Make token ranges optional
- [x] Try removing `ExpandedNameOwned::prefix`
- [x] Remove 4GB input limit
- [ ] ~~Try using custom `ShortRange` for attributes and namespaces indexes. Like 20.12 packed in u32~~
- [x] Test peak memory usage

tomjw64:
:eyes:

tomjw64:
👀 I'll spend a few hours on this.

RazrFalcon:
@tomjw64 Thanks. Removing `ExpandedNameOwned::prefix` is the most important one. It's just garbage. But would require some changes to the parser.

A custom `ShortRange` a bit more tricky. I will probably look into it myself.

RazrFalcon:
Token ranges are surprisingly not that memory-heavy. On the 1GB XML from #18 they produce about 5% memory overhead. They still should be optimized though.

PR #66, opened 2022-10-27
RFC: Track parent prefixes using a stack to drop ExpandedNameOwned::prefix 

adamreichold:
This my stab at the item mentioned in #63 and I have to admit that I am bit out of my depth here meaning that I wrote this based on reading the code and running the tests, not based on a actually understanding the XML processing parts. This is why I added the redundant tracking first, asserting that it would be equivalent to the in-node data and only in the last commit actually removed the `ExpandedOwnedName::prefix` field.

With this in mind, I am sorry if this is completely bogus, especially the second commit as I am unable to argue why there should be no branching, I just did never observe this during the tests.

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -670,7 +670,6 @@ fn process_element<'input>(
                 NodeKind::Element {
                     tag_name: ExpandedNameOwned {
                         ns: tag_ns_uri,
-                        prefix: tag_name.prefix.as_str(),
                         name: tag_name.name.as_str(),
                     },
                     attributes,
@@ -685,14 +684,14 @@ fn process_element<'input>(
             let prefix = prefix.as_str();
             let local = local.as_str();
 
+            let parent_node = &mut doc.nodes[parent_id.get_usize()];
             let parent_prefix = *parent_prefixes.last().unwrap();
 
-            doc.nodes[parent_id.get_usize()].range.end = token_span.end() as u32;
-            if let NodeKind::Element { ref tag_name, .. } = doc.nodes[parent_id.get_usize()].kind {
-                assert_eq!(tag_name.prefix, parent_prefix);
-                if prefix != tag_name.prefix || local != tag_name.name {
+            parent_node.range.end = token_span.end() as u32;
+            if let NodeKind::Element { ref tag_name, .. } = parent_node.kind {
+                if prefix != parent_prefix || local != tag_name.name {
                     return Err(Error::UnexpectedCloseTag {
-                        expected: gen_qname_string(tag_name.prefix, tag_name.name),
+                        expected: gen_qname_string(parent_prefix, tag_name.name),
                         actual: gen_qname_string(prefix, local),
                         pos: err_pos_from_span(doc.text, token_span),
                     });
@@ -700,7 +699,7 @@ fn process_element<'input>(
             }
             pd.awaiting_subtree.push(*parent_id);
 
-            if let Some(id) = doc.nodes[parent_id.get_usize()].parent {
+            if let Some(id) = parent_node.parent {
                 *parent_id = id;
                 parent_prefixes.pop();
                 debug_assert!(!parent_prefixes.is_empty());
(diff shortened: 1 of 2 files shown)


RazrFalcon:
Nice! Can we put `parent_prefixes` in `ParserData`?

Overall, I don't remember how I wrote this code to begin with. So as long as the tests are passing we're fine.

src/parse.rs
```diff
@@ -682,20 +684,24 @@ fn process_element<'input>(
             let prefix = prefix.as_str();
             let local = local.as_str();
 
-            doc.nodes[parent_id.get_usize()].range.end = token_span.end() as u32;
-            if let NodeKind::Element { ref tag_name, .. } = doc.nodes[parent_id.get_usize()].kind {
-                if prefix != tag_name.prefix || local != tag_name.name {
+            let parent_node = &mut doc.nodes[parent_id.get_usize()];
+            let parent_prefix = *parent_prefixes.last().unwrap();
```
RazrFalcon, line 688: Do we guarantee that prefixes are not empty?

adamreichold: I changed the first commit to check

```rust
debug_assert!(!parent_prefixes.is_empty());
```

instead of

```rust
debug_assert!(parent_prefixes.pop().unwrap());
```

which never fired during the test and I do initialize it with `vec![""]` (which would be the prefix of the root node). But I have admittedly no proof besides always pushing one prefix when I update the parent ID as to why this `unwrap` is unreachable.

I could do `unwrap_or("")` instead but I feared that this would hide mistakes on my part?

tomjw64:
In my WIP branch, I've replaced ExpandedNameOwned with:
```
#[derive(Copy, Clone, Debug, PartialEq)]
struct NamespacedName<'input> {
    namespace_idx: Option<NonMaxU32>,
    local_name: &'input str,
}
```
For reference: `size_of::<ExpandedNameOwned>() = 64`, `size_of::<NamespacedName>() = 24`
Ignoring the whole NonMaxU32 thing for now, we can choose to not only rid ourselves of the prefix, but also the ns Cow, and not have any loss of information.
The one larger issue with doing this is that it forces us to follow the same pattern with `Attribute` as with `Node` (i.e. it is a struct that must be cheap to copy that we generate only when needed. For example, we change the struct definition to:
```
pub struct Attribute<'a, 'input: 'a> {
    doc: &'a Document<'input>,
    d: &'a AttributeData<'input>,
}
```
This only turns out to be a breaking change with regards to two things:
1. Attribute lifetimes (does it realistically matter?)
2. Node::attributes must return an iterator instead of a `&[Attribute]

Though it doesn't reduce memory usage as heavily, this approach is more minimal than mine, and better if these two breaking changes are a price too much to pay.

src/parse.rs
RazrFalcon: Oh, I see now. `parent_prefixes` is never empty. Then `unwrap` is fine. Maybe a comment before `unwrap` would be nice.

adamreichold:
> Can we put parent_prefixes in ParserData?

Done

RazrFalcon:
@tomjw64 

>Node::attributes must return an iterator instead of a `&[Attribute]

No-no-no. I want to have only `&[T]` in public API.

RazrFalcon:
@adamreichold Would you mind if I squash commits?

adamreichold:
> @adamreichold Would you mind if I squash commits?

Not at all. The separate commits are mainly meant to convince myself that this works as intended.

src/parse.rs
adamreichold: > Maybe a comment before unwrap would be nice.

Done

RazrFalcon:
@tomjw64 Subsequent changes to this patch are still fine. There is a lot of room for memory usage optimizations. When writing roxmltree originally, I have spent most of the time fighting with the XML itself, especially entities resolving and namespaces. And now we should focus more on memory overhead. But the public API should still be as simple as possible.

RazrFalcon:
Thanks!

PR #67, opened 2022-10-31
Hide token ranges w.r.t. original input behind feature flag.

adamreichold:
Few things I am unsure about:

- Did I find all of them and is a compile-time feature what was intended by #63?
- Should this be a default (would be mostly backwards compatible) or non-default (ecosystem automatically picks up gains) feature?
- Should `Document::text_pos_at`, `Document::input_text` and `Document::input` also be gated behind this?
- The unit tests currently work only with the feature enabled. So maybe a default feature after all or rework the tests to assert ranges only if enabled?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,9 +19,6 @@ exclude = ["test-apps", "testing-tools"]
 xmlparser = "0.13.5"
 
 [features]
-default = ["std"]
+default = ["std", "token-ranges"]
 std = []
 token-ranges = []
-
-[package.metadata.docs.rs]
-features = ["token-ranges"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,3 +21,7 @@ xmlparser = "0.13.5"
 [features]
 default = ["std"]
 std = []
+token-ranges = []
+
+[package.metadata.docs.rs]
+features = ["token-ranges"]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -75,7 +75,7 @@ type Range = core::ops::Range<usize>;
 pub struct Document<'input> {
     /// An original data.
     ///
-    /// Required for `text_pos` methods.
+    /// Required for `text_pos_at` methods.
     text: &'input str,
     nodes: Vec<NodeData<'input>>,
     attrs: Vec<Attribute<'input>>,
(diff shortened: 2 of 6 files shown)


RazrFalcon:
1. Yes, we have one in Node and two in Attribute. You should feature gate `ShortRange` as well.
2. Yes, those methods should be feature-gated as well.
3. Are you sure? I don't think I use token ranges in tests. The only exception is `print_pos` example.

adamreichold:
> 1. Yes, we have one in Node and two in Attribute. You should feature gate `ShortRange` as well.

`ShortRange` is also used in `NodeKind::Element::{attributes, namespaces}`. At least for `namespaces`, it does not look like it can made optional?

>     2. Yes, those methods should be feature-gated as well.

I tried, but at least `Document::text` is used in a lot of calls to `err_pos_from_qname` so this would only remove the methods which seems not really useful since they only need this field?

>     3. Are you sure? I don't think I use token ranges in tests. The only exception is `print_pos` example.

There were a few `test_pos_*` test cases, but I adjust both tests and examples to work out of the box. (I still test `--all-features`  in the CI though.)

Still undiscussed:

> Should this be a default (would be mostly backwards compatible) or non-default (ecosystem automatically picks up gains) feature?

RazrFalcon:
Oh, right `ShortRange` does used.

And yes, I think it should be enabled by default.

adamreichold:
> And yes, I think it should be enabled by default.

Done.

Cargo.toml
```diff
@@ -15,9 +15,20 @@ edition = "2018"
...
 
+[[example]]
+name = "ast"
+
+[[example]]
+name = "stats"
+
+[[example]]
+name = "print_pos"
+required-features = ["token-ranges"]
```
RazrFalcon, line 26: I don't think we need this anymore.

Or at least we should simply feature gate the main in the example file.

src/lib.rs
```diff
@@ -449,6 +452,7 @@ impl<'input> Attribute<'input> {
         &self.value
     }
 
+    #[cfg(feature = "token-ranges")]
```
RazrFalcon, line 455: Please place below the doc comment.

RazrFalcon:
Also update the CI to run tests without this feature.

adamreichold:
> Also update the CI to run tests without this feature.

Done.

src/lib.rs
adamreichold: Done.

Cargo.toml
adamreichold: Without this, `cargo test --no-default-features` would fail. I therefore feature-gated the main function instead of listing the examples.

RazrFalcon:
Thanks!

PR #68, opened 2022-10-31
RFC: Use empty namespace as niche to avoid space overhead of Option in ExpandedNameOwned.

adamreichold:
If XML does allow empty namespaces, then this optimization is not valid. Otherwise, it reduces the size of `ExpandedNameOwned` from 48 to 40 bytes on amd64, i.e. a 16% reduction.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -415,7 +415,7 @@ impl<'input> Attribute<'input> {
     /// ```
     #[inline]
     pub fn namespace(&self) -> Option<&str> {
-        self.name.ns.as_ref().map(Cow::as_ref)
+        if !self.name.ns.is_empty() { Some(&*self.name.ns) } else { None }
     }
 
     /// Returns attribute's name.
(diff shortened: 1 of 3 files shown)


RazrFalcon:
>If XML does allow empty namespaces, then this optimization is not valid. 

Who known. XML namespace are bizarre. 😄 

But since we follow libxml2/lxml, it should be fine:

```xml
<e xmlns:n=''/>
```

```
lxml.etree.XMLSyntaxError: xmlns:n: Empty XML namespace is not allowed, line 1, column 14
```

So this change is fine, but please add a test case.

adamreichold:
> So this change is fine, but please add a test case.

Added a trivial one and this change does alter the externally visible behaviour of the API, i.e. when explicitly specifying an empty namespace, we go from an attribute with an empty namespace to an attribute without a namespace.

I wonder whether `Namespaces::push_ns` should be adjusted to fail for empty `uri` with an explicit error similar to libxml2/lxml?

adamreichold:
> I wonder whether Namespaces::push_ns should be adjusted to fail for empty uri with an explicit error similar to libxml2/lxml?

But then the test "ns_011" seems to explicitly use an empty URI for the default namespace...

RazrFalcon:
You're right. The default namespace `xmlns` can be empty. But a custom one cannot.

I guess we can check that non-`xmlns` namespaces for being non-empty and return and error (if not already), but still return `None` for an empty `xmlns`. Or we could always return a string and just mention that it can be empty and this is the caller problem now.

Either way should be fine.

PR #69, opened 2022-11-01
Use a custom version of Cow<'a, str> that keeps a niche availabel to reduce the size of ExpandedNameOwned.

adamreichold:
This supersedes #68 and achieves the same size reduction of `ExpandedNameOwned` with any behavioural changes.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28,6 +28,7 @@ extern crate alloc;
 extern crate std;
 
 use alloc::borrow::Cow;
+use alloc::boxed::Box;
 use alloc::vec::Vec;
 use core::cmp::Ordering;
 use core::fmt;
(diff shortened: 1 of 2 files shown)


adamreichold:
I think it might make sense to use `CowStr<'a>` everything a `Cow<'a, str>` is currently used in the construction of the tree since it also sheds excess capacity which seems reasonable for a read-only data structure even though the extra niche might go unused.

(Just as a heads up for me being unresponsive: I will not be able to work on this again until the next weekend.)

src/lib.rs
```diff
@@ -311,6 +312,32 @@ impl ShortRange {
     }
 }
 
+/// A custom variant of `Cow<'a, str>` that
+/// keeps a niche available and sheds excess capacity.
+#[derive(Clone, PartialEq)]
+enum CowStr<'a> {
+    Borrowed(&'a str),
+    Owned(Box<str>),
```
RazrFalcon, line 320: What kind of magic is this? Is this a `String` somehow? I never seen this trick before.

Also, why it's smaller than `Cow` exactly? I think it worth documenting this as well.

adamreichold: > What kind of magic is this? Is this a String somehow? I never seen this trick before.

`Box<str>` is to `String` as `Box<[T]>` is to `Vec<T>`, i.e. it is a boxed slice meaning a two-word large fat pointer into a heap allocation.

> Also, why it's smaller than Cow exactly? I think it worth documenting this as well.

Actually, `CowStr<'a>` is not smaller than `Cow<'a, str>`, both a 24 bytes on amd64 but this only due to `Cow<'a, str>` using the niche of `&'a str` and `String` (both are never all zeros as even `String::new` puts a non-zero dangling pointer in there) to store its tag. (Naively, it would be 32 bytes large, i.e. tag, point, length and capacity.)

But `CowStr<'a>` is 24 bytes (tag, pointer and length) without using the niche optimization itself so that it remains available for a situation like `Option<CowStr<'a>>` where the tag of `Option` can be stored in the niche of `CowStr<'a>` so that `size_of::<Option<CowStr<'a>>>() == size_of::<CowStr<'a>>()`.

Shedding excess capacity in this case refers to `Box<str>` not having a capacity and hence storing only exactly the number of bytes required to hold the string. This does have the downside that going from `String` to `Box<str>` might re-allocate to actually get an allocation without excess capacity at the end.

Will try to expand on the comment on the weekend...

adamreichold: > This does have the downside that going from String to Box<str> might re-allocate to actually get an allocation without excess capacity at the end.

Maybe just adding that this is not particularly likely as allocators tend to lie anyway, e.g. they round up to size classes to satisfy small allocations and while this then can still result in a call to `realloc`, that call will return the same allocation if it stays within the same size class.

adamreichold: > `Box<str>` is to `String` as `Box<[T]>` is to `Vec<T>`, i.e. it is a boxed slice meaning a two-word large fat pointer into a heap allocation.

To expand on this: The point is that it cannot be expanded without re-allocation in contrast to `Vec` with keeps some uninitialized capacity at the end and needs to re-allocate rarely (amortized by doubling the capacity etc.).

adamreichold: Expanded on the comment a bit. Not sure if you want more details on the difference between `Box<str>` and `String` though.

RazrFalcon: So I did some research to figure out what `Box<str>`  is and I think I got it. I simply never heard of this trick. Very useful for out use case.

Issue #18, opened 2019-11-15
How to Read 2 GB file and Parse it


vnnv01:
Hi Team,

Lxml in python is able to read only an element and we can remove it from memory after we parse it. This helps in parsing big file easily without affecting memory.  It has method called iterparse

https://www.ibm.com/developerworks/xml/library/x-hiperfparse/

Is it possible to have something similar , so that we dont need to create a huge string and parse it.

Thanks

RazrFalcon:
Have you considered `memmap`?

`roxmltree` designed for small files in mind, so I'm not sure what exactly you want to change to support large files. Also, I have no idea how lxml handles large files.

The core idea of `roxmltree` is to reduce the number of allocations by reusing an input string as much as possible. So we still need a whole string.

dabreegster:
Sorry to reopen an old issue, but I'm trying to handle a 1GB input. I'm running out of memory (16GB) trying to do this in the naive way.

For starters, I'm not sure I understand why -- a 1GB utf-8 string shouldn't take so much RAM, since [the representation](https://doc.rust-lang.org/std/string/struct.String.html#representation) is just raw bytes.

I'm trying memmap:

```
let f = std::fs::File::open(path)?;
let mmap = unsafe { memmap2::Mmap::map(&f)? };
let x = std::str::from_utf8(&mmap)?;
let doc = roxmltree::Document::parse(x)?;
```

but still OOMing at `str::from_utf8`. This isn't an issue with roxmltree, but in case you have any ideas, it's much appreciated. As you've said here and in #62, this crate is meant for small inputs, so I maybe need to just switch to a streaming reader.

RazrFalcon:
Can you send me the file?

I haven't measured roxmltree overhead, but it should be around 3-4x of the original file. Not good, but not 16GB either.

Ans since it crashes before roxmltree, I would assume something wrong with your file or hardware.

dabreegster:
https://www.dropbox.com/s/2qqv3s5nxs8odi7/greater_london.osm?dl=0
I'm mistaken, reading to a string is fine. It is OOMing during `Document::parse`. This might just be a matter of #63.

RazrFalcon:
Can confirm. It takes 12GB to parse in Release mode. Meaning a 12x overhead. I guess you should wait for #63

PR #71, opened 2022-11-07
Deduplicate tag and attribute names.

adamreichold:
I am not even sure this is serious as this massively trades off runtime (O(log n) search time and indirection behind `Rc`) for space.

It does reduce peak memory consumption from 7.8 GB to 5.5 GB for the document from #18 though (both of which include the 1.1 GB of reading the document itself though). Interestingly enough, `doc.names` has 17 entries in this case.

I think the search time could be reduced by using a hash table but that complicates `no_std` builds. Furthermore, both hash tables without keyed hashes and sorted slices are vulnerable to quadratic complexity DoS attacks (e.g. forcing insertion at the beginning for each parsed name).

Finally, the additional overhead of `Rc` would always be there for small documents and this opens up the additional question of using `Rc` versus `Arc`, i.e. whether `Document: Send` holds or not.

In any case, I wanted to post this for discussion even if I don't think this is really viable. But it does shed some light on the possible savings with regards to deduplication via indexes as proposed by @tomjw64 where `&[Attribute]` would have to become `impl ExactSizeIterator<Item=Attribute>` in the API.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -30,6 +30,7 @@ extern crate std;
 use alloc::borrow::Cow;
 use alloc::boxed::Box;
 use alloc::vec::Vec;
+use alloc::rc::Rc;
 use core::cmp::Ordering;
 use core::fmt;
 use core::hash::{Hash, Hasher};
@@ -81,6 +82,8 @@ pub struct Document<'input> {
     nodes: Vec<NodeData<'input>>,
     attrs: Vec<Attribute<'input>>,
     namespaces: Namespaces<'input>,
+    /// Sorted list of deduplicated tag and attribute names
+    names: Vec<Rc<ExpandedNameOwned<'input>>>,
 }
 
 impl<'input> Document<'input> {
@@ -334,7 +337,7 @@ impl ShortRange {
 ///    when an owned string had to be constructed thereby
 ///    keeping only the amount of memory around that is
 ///    required to store the string.
-#[derive(Clone, PartialEq)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
 enum CowStr<'a> {
     Borrowed(&'a str),
     Owned(Box<str>),
@@ -415,7 +418,7 @@ impl From<usize> for NodeId {
 enum NodeKind<'input> {
     Root,
     Element {
-        tag_name: ExpandedNameOwned<'input>,
+        tag_name: Rc<ExpandedNameOwned<'input>>,
         attributes: ShortRange,
         namespaces: ShortRange,
     },
@@ -439,7 +442,7 @@ struct NodeData<'input> {
 /// An attribute.
 #[derive(Clone)]
 pub struct Attribute<'input> {
-    name: ExpandedNameOwned<'input>,
+    name: Rc<ExpandedNameOwned<'input>>,
     value: Cow<'input, str>,
     #[cfg(feature = "token-ranges")]
     range: ShortRange,
@@ -624,7 +627,7 @@ impl<'input> Deref for Namespaces<'input> {
 }
 
 
-#[derive(Clone, PartialEq)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
 struct ExpandedNameOwned<'input> {
     ns: Option<CowStr<'input>>,
     name: &'input str,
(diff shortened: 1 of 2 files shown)


RazrFalcon:
Yes, I though about this, but I would assume performance penalty isn't worth it. Have you tried running the benchmarks?
Also, I'm not sure we even need `Rc` here. A simple index would be just fine.

As for the HashTable, it should be fine. roxmltree is no_std but it still uses `alloc`. So we have an access to HashTable.

To speed up the process we could use something like this:

```rust
fn string_hash(s: &str) -> u64 {
    let mut h = alloc::collections::hash_map::DefaultHasher::new();
    s.hash(&mut h);
    h.finish()
}

// and then

names: Vec<ExpandedNameOwned<'input>>, // stored
namesHashes: Vec<u64>, // only during parsing, sorted(?)
```

Basically a very specific kind of HashTable. 
Or even uses a custom, faster hash method. One additional dependency should be fine.

Anyway, this is a good idea as long as the performance penalty isn't that bad.

As for `Send` - I don't really care.

adamreichold:
> As for the HashTable, it should be fine. roxmltree is no_std but it still uses alloc. So we have an access to HashTable.

As you can see at https://doc.rust-lang.org/stable/alloc/collections/index.html, the `alloc` crate does not contain a `HashMap`. This is most likely because `DefaultHasher` calls into the OS for entropy (e.g. via `getrandom`) which `alloc` cannot do. The `hashbrown` crate can be used on `no_std` though (without a runtime random key).

> Also, I'm not sure we even need Rc here. A simple index would be just fine.

I think this is the route described by

> But it does shed some light on the possible savings with regards to deduplication via indexes as proposed by @tomjw64 where &[Attribute] would have to become impl ExactSizeIterator<Item=Attribute> in the API.

meaning that e.g. an `Attribute` cannot produce its name without access to the `Document` so that it either needs a facade or the user has to manually pass `&Document` where the name is needed.

> Basically a very specific kind of HashTable.

Of course, the snippet is missing how collisions are handled which is sort of the thing which makes the difference between a good and a bad hash table.

> Anyway, this is a good idea as long as the performance penalty isn't that bad.

I think the index-based version would be nice since it does not open up the `Send` question and avoids the extra indirection in storing the `ExpandedNameOwned` structs and I would go down the benchmarks rabbit hole for that one but you seem pretty strongly opposed to that in https://github.com/RazrFalcon/roxmltree/pull/66#issuecomment-1294006135 and the following?

RazrFalcon:
>the alloc crate does not contain a HashMap

Wow, I didn't know that.

>I think this is the route described by

Oh, my bad. 

>Of course, the snippet is missing how collisions are handled

Do you think we would be able to trigger a hash collision? I'm not an expert in this area. 
I cannot see this happening on a typical file, but someone may construct a bad one...
From my understanding, a hash collision resolving would basically require a `==` call after comparing the hash, right?

>I think the index-based version would be nice since

So an iterator type that stores `&Document` and produces `Attribute`? I guess we have no other choice...

Either way, the memory impact is too good to ignore, but performance is important as well.
We either use more memory or check every single element and attribute name for uniqueness, which seem like a performance nightmare to me.

adamreichold:
> Either way, the memory impact is too good to ignore, but performance is important as well.
We either use more memory or check every single element and attribute name for uniqueness, which seem like a performance nightmare to me.

Yes, the impact is significant and the index-based version does not come the caveats of the `Rc`-based one, so I think that one is worth a try.

I will hopefully be able to dive into it this weekend. Or maybe @tomjw64 can propose something while I take care of benchmarking #73 now that the

> So an iterator type that stores &Document and produces Attribute? I guess we have no other choice...

is on the table again?

RazrFalcon:
>is on the table again?

For such a big improvement - sure.

And we could replace `namespace()` with an iterator as well then. Removing `Rc` completely.

adamreichold:
Just for completeness, this is the parsing impact of the sorted-list reference-counting version presented here:

```console
> cargo benchcmp master dedup-names 
 name              master ns/iter  dedup-names ns/iter  diff ns/iter  diff %  speedup 
 large_roxmltree   3,261,437       4,550,550               1,289,113  39.53%   x 0.72 
 medium_roxmltree  769,474         1,027,545                 258,071  33.54%   x 0.75 
 tiny_roxmltree    4,981           5,829                         848  17.02%   x 0.85 
```

Pretty gruesome, I'd say. But I'd also hope that the index-based version fears better, especially if coupled with a hash table instead of a sorted list.

RazrFalcon:
Yes, this is expected. roxmltree is tailored towards smaller files were the current logic works just fine. You cannot have both.

Hopefully there are ways to improve uniqueness check performance.

PR #73, opened 2022-11-07
Use std's pattern matching functions instead of byte-wise comparisons.

adamreichold:


src/parse.rs
```diff
@@ -459,8 +459,8 @@ fn parse(text: &str, opt: ParsingOptions) -> Result<Document, Error> {
     };
 
     // Trying to guess rough nodes and attributes amount.
-    let nodes_capacity = text.bytes().filter(|c| *c == b'<').count();
-    let attributes_capacity = text.bytes().filter(|c| *c == b'=').count();
+    let nodes_capacity = text.match_indices('<').count();
+    let attributes_capacity = text.match_indices('=').count();
```
tomjw64, line 463: Curious, why `match_indices` instead of `matches`?

RazrFalcon:
Have you run the benchmarks? Is it faster?

src/parse.rs
adamreichold: No reason besides looking through the methods and not seeing `matches`. Will change.

adamreichold:
> Have you run the benchmarks? Is it faster?

Not yet. Will do, probably this evening.

RazrFalcon:
Ok. I'm just confused by this change. Is it purely stylistic or actually has some impact on parsing.

adamreichold:
So it is consistently slower:

```console
> cargo benchcmp master pattern
 name              master ns/iter  pattern ns/iter  diff ns/iter  diff %  speedup 
 large_roxmltree   3,261,437       4,048,171             786,734  24.12%   x 0.81 
 medium_roxmltree  769,474         855,463                85,989  11.18%   x 0.90 
 tiny_roxmltree    4,981           5,532                     551  11.06%   x 0.90 
```

which I was _not_ expecting. Thank you calling out the missing benchmarks.

(I had assumed that std would be clever enough to use libc's optimized implementations of `memchr` and/or `memmem` but that does not seem to be the case.)

So I definitely withdraw this. Sorry for the noise!

RazrFalcon:
No problem! That's why we have benchmarks.

I remember testing this to get as best results as possible without relying on `memchr`, since I don't want an additional dependency.

Also, afair, the `memchr` crate uses SIMD only on x86. So on ARM it will use the same code as we do already.

PR #74, opened 2022-11-08
Dedup attribute names using an IndexSet

adamreichold:
This could be a first step towards what we discussed in #73 but things are ... complicated.

This alone does improve the memory consumption for the `greater_london.osm` file from 7.8 GB to 6.3 GB.

However, using the default configuration of the `indexmap` crate, performance is significantly degraded and more importantly fluctuates widely depending on the randomly chosen SipHash key, e.g. for a particular run I get

```console
> cargo benchcmp master dedup-attrs
 name              master ns/iter  dedup-attrs ns/iter  diff ns/iter  diff %  speedup 
 large_roxmltree   3,548,058       4,118,748                 570,690  16.08%   x 0.86 
 medium_roxmltree  775,340         987,012                   211,672  27.30%   x 0.79 
 tiny_roxmltree    4,987           5,611                         624  12.51%   x 0.89 
```

but I also get runs which measurably faster or slower (which I do not get on the `main` branch with CPU frequency boost disabled and using the `performance` CPU governor).

When I use the `ahash` crate to replace SipHash by AHash, things look somewhat better, especially for the large test case:

```console
> cargo benchcmp master dedup-attrs-ahash 
 name              master ns/iter  dedup-attrs-ahash ns/iter  diff ns/iter  diff %  speedup 
 large_roxmltree   3,548,058       3,653,333                       105,275   2.97%   x 0.97 
 medium_roxmltree  775,340         924,645                         149,305  19.26%   x 0.84 
 tiny_roxmltree    4,987           5,569                               582  11.67%   x 0.90 
```

But while this still enables `no_std` builds, it somewhat complicates the MSRV build...


src/parse.rs
```diff
@@ -799,13 +800,13 @@ fn resolve_attributes<'input>(
         let attr_name = ExpandedNameOwned { ns, name: attr.local.as_str() };
```
adamreichold, line 800: It is somewhat pointless to possible allocate `ns` here only to throw it away again after deduplication, but I did not find the strength to shave this particular yak yet...

adamreichold:
Having written this code, but taking a step back, I think we should _not_ do this. Personally, I think its simplicity is one of the main selling points of `roxmltree` and this makes things significantly more complicated (especially in the software engineering department, i.e. managing dependencies and their configuration) while also slowing down the typical use case of small files (which is actually the use I am personally interested in).

tomjw64:
Looks like we had a lot of the same ideas: https://github.com/RazrFalcon/roxmltree/pull/75

RazrFalcon:
Two new dependencies? And such a big one as `indexmap`? No-no-no. This crate should stay dependency free.

Also, the MSRV hack is too much.

As you have said, this crate is all about simplicity. Well, except inherited complexity thanks to XML itself.

adamreichold:
> Two new dependencies? And such a big one as indexmap? No-no-no. This crate should stay dependency free.

Fair enough.

PR #75, opened 2022-11-08
Breaking Change: Use indexes for Node and Attribute Namespaces

tomjw64:
This change is breaking because the public struct `Attributes` would have a different number of lifetime arguments and `Node::attributes` would return an iterator instead of a slice.

This might represent a good starting place. @adamreichold seems to be trying to find ways to intern names, which could bring a huge benefit as well. I don't think this approach would conflict with that strategy.

Note: NamespacedName is 24 bytes instead of 20 by attempting to utilize `Option<NonZeroU32>` because the alignment of `&str` is 8.
```
MEMORY SIZES BEFORE
-------------
size_of::<NodeKind>() = 64
size_of::<NodeData>() = 88
size_of::<ExpandedNameOwned>() = 40
size_of::<Attribute>() = 88
size_of::<Namespace>() = 48

MEMORY SIZES AFTER
------------
size_of::<NodeKind>() = 48
size_of::<NodeData>() = 72
size_of::<NamespacedName>() = 24
size_of::<AttributeData>() = 72
size_of::<Attribute>() = 16
size_of::<Namespace>() = 48

PEAK MEMORY BEFORE
------------------
1.36GB on 100 MB document

PEAK MEMORY AFTER
------------------
1.14GB on 100 MB document

BENCHMARKS
----------
name                                    master ns/iter  idx-for-namespace ns/iter  diff ns/iter  diff %  speedup 
large_roxmltree                         3,062,501       2,820,317                      -242,184  -7.91%   x 1.09 
medium_roxmltree                        640,630         657,078                          16,448   2.57%   x 0.97 
tiny_roxmltree                          4,287           4,136                              -151  -3.52%   x 1.04 
roxmltree_iter_children                 2,306           2,166                              -140  -6.07%   x 1.06 
roxmltree_iter_descendants_expensive    481,111         539,422                          58,311  12.12%   x 0.89
roxmltree_iter_descendants_inexpensive  32,232          29,908                           -2,324  -7.21%   x 1.08 
```

roxml-mem-diff-2022-10-08

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28,7 +28,6 @@ extern crate alloc;
 extern crate std;
 
 use alloc::borrow::Cow;
-use alloc::boxed::Box;
 use alloc::vec::Vec;
 use core::cmp::Ordering;
 use core::fmt;
@@ -44,9 +43,13 @@ pub use crate::parse::*;
 
 /// The <http://www.w3.org/XML/1998/namespace> URI.
 pub const NS_XML_URI: &str = "http://www.w3.org/XML/1998/namespace";
+/// The prefix 'xml', which is by definition bound to NS_XML_URI
+const NS_XML_PREFIX: &str = "xml";
 
 /// The <http://www.w3.org/2000/xmlns/> URI.
 pub const NS_XMLNS_URI: &str = "http://www.w3.org/2000/xmlns/";
+/// The string 'xmlns', which is used to declare new namespaces
+const XMLNS: &str = "xmlns";
 
 
 type Range = core::ops::Range<usize>;
@@ -124,6 +127,7 @@ impl<'input> Document<'input> {
 
     fn get_attribute<'a>(&'a self, idx: usize) -> Attribute<'a, 'input> {
         Attribute {
+            doc: self,
             d: &self.attrs[idx],
         }
     }
@@ -326,53 +330,6 @@ impl ShortRange {
     }
 }
 
-/// A custom variant of `Cow<'a, str>` that
-/// keeps a niche available and sheds excess capacity.
-///
-/// This variant stores `Box<str>` instead of `String`
-/// which is more appropriate for a space-conscious
-/// read-only data structure as
-///
-/// 1. it keeps a niche available so that for example
-///
-///    ```rust,ignore
-///    assert_eq!(
-///        size_of::<CowStr<'static>>(),
-///        size_of::<Option<CowStr<'static>>>(),
-///    );
-///    ```
-///
-///    holds and
-///
-/// 2. it sheds any excess capacity that was allocated
-///    when an owned string had to be constructed thereby
-///    keeping only the amount of memory around that is
-///    required to store the string.
-#[derive(Clone, PartialEq)]
-enum CowStr<'a> {
-    Borrowed(&'a str),
-    Owned(Box<str>),
-}
-
-impl<'a> From<Cow<'a, str>> for CowStr<'a> {
-    fn from(v: Cow<'a, str>) -> CowStr<'a> {
-        match v {
-            Cow::Borrowed(v) => CowStr::Borrowed(v),
-            Cow::Owned(v) => CowStr::Owned(v.into()),
-        }
-    }
-}
-
-impl<'a> AsRef<str> for CowStr<'a> {
-    fn as_ref(&self) -> &str {
-        match self {
-            CowStr::Borrowed(v) => v,
-            CowStr::Owned(ref v) => v,
-        }
-    }
-}
-
-
 /// A node ID stored as `u32`.
 ///
 /// An index into a `Tree`-internal `Vec`.
@@ -451,7 +408,7 @@ struct NodeData<'input> {
 
 #[derive(Clone)]
 struct AttributeData<'input> {
-    name: ExpandedNameOwned<'input>,
+    name: NamespacedName<'input>,
     value: Cow<'input, str>,
     #[cfg(feature = "token-ranges")]
     range: ShortRange,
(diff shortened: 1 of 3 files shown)


adamreichold (commented):
I think embedding `AttributeData` inside `Attribute` is much nicer than what I did in #74 as it results in much less duplication.

I think the iterator `Attributes` could be simpler by relying on `core::slice::Iter` though, thereby also removing the need for `Node::attribute_data` and `Document::get_attribute` as well as some bounds checking.

Finally, if we do decide to go for #74 in addition to this, we probably want to use an `IndexSet` for `Document::namespaces` as well instead of the linear search employed now.

src/lib.rs
```diff
@@ -435,19 +406,24 @@ struct NodeData<'input> {
...
     #[cfg(feature = "token-ranges")]
     range: ShortRange,
     #[cfg(feature = "token-ranges")]
     value_range: ShortRange,
 }
 
-impl<'input> Attribute<'input> {
+/// An attribute.
+#[derive(Copy, Clone)]
+pub struct Attribute<'doc, 'input: 'doc> {
```
RazrFalcon, line 421: I think we should stick to `'a` instead of `'doc` for consistency.

src/lib.rs
```diff
@@ -312,53 +330,6 @@ impl ShortRange {
     }
 }
 
-/// A custom variant of `Cow<'a, str>` that
```
RazrFalcon, line 315: Are you sure we don't need this anymore?

src/parse.rs
```diff
@@ -1075,42 +1075,41 @@ fn _normalize_attribute(
...
-                // If an URI was not found and prefix IS empty than
-                // an element or an attribute doesn't have a namespace.
-                //
-                // Example:
-                // <e a='b'/>
-                Ok(None)
-            }
-        }
-    }
+    range.to_urange().into_iter()
```
RazrFalcon, line 1089: Can we make it less functional? It's hard to read.

src/lib.rs
adamreichold: `ExpandedNameOwned` is gone and was the only user. Applying it elsewhere does not have clear benefits (I tried).

RazrFalcon: Oh, I see. That was a nice trick.

RazrFalcon:
Can you explain the new logic/algorithm a bit? It's hard to figure out from the patch itself.

src/lib.rs
```diff
@@ -44,9 +43,13 @@ pub use crate::parse::*;
 
 /// The <http://www.w3.org/XML/1998/namespace> URI.
 pub const NS_XML_URI: &str = "http://www.w3.org/XML/1998/namespace";
+/// The prefix 'xml', which is by definition bound to NS_XML_URI
```
RazrFalcon, line 46: I think we should leave this change to the next patch.

src/lib.rs
```diff
@@ -435,19 +406,24 @@ struct NodeData<'input> {
...
     #[cfg(feature = "token-ranges")]
     value_range: ShortRange,
 }
 
-impl<'input> Attribute<'input> {
+/// An attribute.
+#[derive(Copy, Clone)]
+pub struct Attribute<'doc, 'input: 'doc> {
+    doc: &'doc Document<'input>,
+    d: &'doc AttributeData<'input>,
```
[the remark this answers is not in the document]
RazrFalcon: Agree.

src/lib.rs
RazrFalcon: Can't we use it for `AttributeData.value`?

src/lib.rs
```diff
@@ -623,33 +599,27 @@ impl<'input> Deref for Namespaces<'input> {
     }
 }
 
-
-#[derive(Clone, PartialEq)]
-struct ExpandedNameOwned<'input> {
-    ns: Option<CowStr<'input>>,
-    name: &'input str,
+#[derive(Copy, Clone)]
+struct NamespacedName<'input> {
```
RazrFalcon, line 603: I think we should keep calling it `ExpandedNameSomething`, otherwise its a bit confusing.

src/lib.rs
adamreichold: We could, but both `Cow<'input, str>` and `CowStr<'input>` are 24 bytes on their own, so the struct itself would not get smaller. (Only `Option<CowStr<'input>>` is smaller than `Option<Cow<'input, str>>`.)

The main difference would be that we should any excess capacity of the owned string, but due to allocator size classes this does not actually save memory most of the time. So I think this alone is not worth the trouble of having a bespoke `Cow` variant.

RazrFalcon: Go it.

src/lib.rs
```diff
@@ -1289,6 +1268,73 @@ impl<'a, 'input: 'a> fmt::Debug for Node<'a, 'input> {
...
+            let next = self.doc.get_attribute(self.current);
+            self.current += 1;
+            Some(next)
+        } else {
+            None
+        }
+    }
+
+    #[inline]
+    fn nth(&mut self, n: usize) -> Option<Self::Item> {
```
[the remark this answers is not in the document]
adamreichold: Actually, this is not good advice as the manual `fold` implementation is not appreciably more efficient than the default one.

src/lib.rs
```diff
@@ -1289,6 +1268,73 @@ impl<'a, 'input: 'a> fmt::Debug for Node<'a, 'input> {
...
+            let next = self.doc.get_attribute(self.until);
+            self.until -= 1;
+            Some(next)
+        } else {
+            None
+        }
+    }
+}
+
+impl ExactSizeIterator for Attributes<'_, '_> {}
```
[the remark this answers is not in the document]
adamreichold: Since this is purely cosmetic, I'd say we drop this and trust in the optimizer removing the asserts.

tomjw64:
Closing in favor of  #76

PR #76, opened 2022-11-17
Follow-up: Use indexes for Node and Attribute Namespaces

adamreichold:
I am sorry if this is considered invasive, but since I would really like to see #75 in a release and it seemed like only cosmetics were left to do and I had some time on my hands, I went through the open items in #75 and tried to address them by amending or adding commits from the original PR.

As to the requested explanation: Instead of storing the namespace URI itself (possibly owned or as a reference into the input) within tags and attributes, this stores only the index and resolves it on-demand into the (already deduplicated) `Document::namespaces` list using the `Attribute` wrapper around `AttributeData`.

For `greater_londom.osm`, this changes reduces the peak memory consumption from 7.8 GB to 6.7 GB.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28,7 +28,6 @@ extern crate alloc;
 extern crate std;
 
 use alloc::borrow::Cow;
-use alloc::boxed::Box;
 use alloc::vec::Vec;
 use core::cmp::Ordering;
 use core::fmt;
@@ -128,6 +127,7 @@ impl<'input> Document<'input> {
 
     fn get_attribute<'a>(&'a self, idx: usize) -> Attribute<'a, 'input> {
         Attribute {
+            doc: self,
             d: &self.attrs[idx],
         }
     }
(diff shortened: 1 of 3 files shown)


RazrFalcon:
@tomjw64 any thoughts?

src/lib.rs
```diff
@@ -28,13 +28,13 @@ extern crate alloc;
...
 
 use alloc::borrow::Cow;
-use alloc::boxed::Box;
 use alloc::vec::Vec;
 use core::cmp::Ordering;
 use core::fmt;
 use core::hash::{Hash, Hasher};
 use core::num::NonZeroU32;
 use core::ops::Deref;
+use core::slice::Iter as SliceIter;
```
RazrFalcon, line 37: Why we need a rename here?

adamreichold: We do not strictly need it, but `Iter` is a very generic name which I would have trouble to place seeing it with the `use` clause. Since it is used only in the definition `struct Attributes`, we could also fully qualify it as `core::slice::Iter` there.

RazrFalcon: >we could also fully qualify it as core::slice::Iter there

Let's do it.

adamreichold: Done.

tomjw64:
> I am sorry if this is considered invasive

Not a problem! I've been busy with other things and I appreciate the additional effort you've put in here :slightly_smiling_face:

tomjw64 (approved):

tomjw64:
I will add that, since some additional changes were made, it would be nice to re-check the parsing benches just to be sure no unintended regressions occurred.

RazrFalcon:
Benchmarks are looking good:

```
 name                                    master ns/iter  devel ns/iter  diff ns/iter   diff %  speedup 
 large_roxmltree                         1,536,984       1,538,130             1,146    0.07%   x 1.00 
 medium_roxmltree                        429,330         427,922              -1,408   -0.33%   x 1.00 
 roxmltree_iter_children                 1,401           1,375                   -26   -1.86%   x 1.02 
 roxmltree_iter_descendants_expensive    565,872         565,008                -864   -0.15%   x 1.00 
 roxmltree_iter_descendants_inexpensive  25,688          22,465               -3,223  -12.55%   x 1.14 
 tiny_roxmltree                          2,196           2,190                    -6   -0.27%   x 1.00
```

PR #78, opened 2022-11-20
Drop attrs_start_idx and use drain when resolving attributes.

adamreichold:


adamreichold:
I had a call `doc.attrs.reserve` in front of the for loop in here but noticed that this should already be taken care of by the `attributes_capacity` computation and would therefore be unnecessary work in the common case.

PR #80, opened 2022-11-21
RFC: Deduplicate namespaces

adamreichold:
While we do not expand namespaces into attributes any more, we still store redundantly expanded namespaces as we traverse the document tree during parsing. Now that we have the interface that puts namespaces behind an indirection, we fully deduplicate them using a sorted list of indexes.

This does not improve memory consumption on `greater_london.osm` as that document does not use any namespaces, but a for a [CSW response](https://en.wikipedia.org/wiki/Catalogue_Service_for_the_Web) of 25 MB I had lying around, this does decrease memory consumption from 120 MB to 108 MB, i.e. by almost 10%.

The benchmarks are actually a mixed bag instead of regressions throughout, most likely due to the improved locality of a smaller working set:

```
 name                                    master ns/iter  dedup-namespaces ns/iter  diff ns/iter   diff %  speedup 
 large_roxmltree                         4,020,847       3,431,061                     -589,786  -14.67%   x 1.17 
 medium_roxmltree                        804,519         810,177                          5,658    0.70%   x 0.99 
 roxmltree_iter_children                 3,586           3,574                              -12   -0.33%   x 1.00 
 roxmltree_iter_descendants_expensive    128,694         128,552                           -142   -0.11%   x 1.00 
 roxmltree_iter_descendants_inexpensive  44,570          44,731                             161    0.36%   x 1.00 
 tiny_roxmltree                          4,860           4,918                               58    1.19%   x 0.99 
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -568,28 +568,29 @@ struct Namespaces<'input> {
 
 impl<'input> Namespaces<'input> {
     #[inline]
-    fn push_ns(&mut self, name: Option<&'input str>, uri: Cow<'input, str>) {
+    fn push_ns(&mut self, name: Option<&'input str>, uri: Cow<'input, str>) -> Result<(), Error> {
         let value = Namespace { name, uri };
         debug_assert_ne!(value.name, Some(""));
 
-        match self
+        let idx = match self
             .sorted
             .binary_search_by(|idx| self.values[*idx as usize].cmp(&value))
         {
-            Ok(sorted_idx) => {
-                let idx = self.sorted[sorted_idx];
-
-                self.tree.push(idx);
-            }
+            Ok(sorted_idx) => self.sorted[sorted_idx],
             Err(sorted_idx) => {
+                if self.values.len() > core::u16::MAX as usize {
+                    return Err(Error::NamespacesLimitReached);
+                }
                 let idx = self.values.len() as u16;
                 self.values.push(value);
-
                 self.sorted.insert(sorted_idx, idx);
-
-                self.tree.push(idx);
+                idx
             }
-        }
+        };
+
+        self.tree.push(idx);
+
+        Ok(())
     }
 
     #[inline]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -508,7 +508,7 @@ impl fmt::Debug for Attribute<'_, '_> {
 /// A namespace.
 ///
 /// Contains URI and *prefix* pair.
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -1003,13 +1035,19 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
     /// assert_eq!(doc.root_element().namespaces().len(), 1);
     /// ```
     #[inline]
-    pub fn namespaces(&self) -> &'a [Namespace<'input>] {
-        match self.d.kind {
+    pub fn namespaces(&self) -> impl ExactSizeIterator<Item = &'a Namespace<'input>> + fmt::Debug {
```
adamreichold, line 1038: I we do this, this should become a named type as having unnameable types in the API of a library is a discourtesy to its users.

adamreichold:
While polishing this I noticed two things:

* Now that the namespaces are deduplicated, there is probably no point in supporting more than 2^16 of them per document.
* We still allocated the normalized namespace URI even though they were thrown away for duplicates which I fixed.

The benchmarks somehow look worse now, but that is because `master` got magically faster even though the benchmarking setup is exactly the same:

```
 name                                    master ns/iter  dedup-namespaces ns/iter  diff ns/iter   diff %  speedup 
 large_roxmltree                         4,222,511       3,525,434                     -697,077  -16.51%   x 1.20 
 medium_roxmltree                        755,132         800,921                         45,789    6.06%   x 0.94 
 roxmltree_iter_children                 3,590           3,745                              155    4.32%   x 0.96 
 roxmltree_iter_descendants_expensive    129,573         118,288                        -11,285   -8.71%   x 1.10 
 roxmltree_iter_descendants_inexpensive  44,061          40,630                          -3,431   -7.79%   x 1.08 
 tiny_roxmltree                          4,766           5,016                              250    5.25%   x 0.95
```

Not sure what to make of that, but especially the medium benchmark seems to jump between 750k and 800k ns/iter depending on codegen whims of rustc, irrespective of which branch I am measuring.

src/lib.rs
adamreichold: We will be able to save a ton of code when https://github.com/rust-lang/rust/issues/63063 gets stabilized and comes within the MSRV reach of this crate. Until then, I added an explicit type.

I also did some boy scouting on `Descendants` to remove some redundant bounds check which I guess shows in the benchmarks below.

adamreichold:
I also tried an adaptive search strategy (linear search up to N values, then binary search), but it did not help. I suspect that the loss of code density was too big and ate into the better temporal locality.

PR #86, opened 2023-01-06
Detect unclosed root nodes

### LINKED ISSUE #85: roxmltree accepts opening tag without closing tag
If I try this:

```rust
Document::parse("<open>Text").unwrap();
```

roxmltree accepts this without complaint. Is this an explicit feature or should it check that there are unclosed opening tags at the end of parsing and reject the input as invalid XML if so?

adamreichold:
The detection methods feels a bit hacky to me but it was already available state which should be sufficient to detect the issue.

Closes #85

### DIFF
--- a/benches/xml.rs
+++ b/benches/xml.rs
@@ -399,12 +399,7 @@ benchmark_group!(
     large_sdx_document,
     huge_sdx_document,
 );
-benchmark_group!(
-    minidom,
-    tiny_minidom,
-    medium_minidom,
-    large_minidom,
-);
+benchmark_group!(minidom, tiny_minidom, medium_minidom, large_minidom,);
 benchmark_group!(
     xmlparser,
     tiny_xmlparser,
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -80,6 +80,9 @@ pub enum Error {
     /// The XML document must have at least one element.
     NoRootNode,
 
+    /// The root node was opened but never closed.
+    UnclosedRootNode,
+
     /// An XML with DTD detected.
     ///
     /// This error will be emitted only when `ParsingOptions::allow_dtd` is set to `false`.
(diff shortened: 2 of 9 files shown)


RazrFalcon:
Thanks!

adamreichold:
I would be glad if you could push out the namespace deduplication and this fix as a 0.16.1 point release? Thank you!

RazrFalcon:
This commit adds a new error, so it cannot be a patch bump. I can release it as 0.17 right now.

adamreichold:
> This commit adds a new error, so it cannot be a patch bump. I can release it as 0.17 right now.

Ah yes, I did not consider this. 0.17.0 would of course be great too.

Adding `#[non_exhaustive]` to the `Error` enum might be nice so that this would not breaking for future release, but IIRC this would bump MSRV to 1.40.

There are workarounds like adding an unmatchable variant like `_NonExhaustive(Private)` where `pub struct Private;` is not re-exported from `mod parse` but that is pretty ugly and the diagnostics are worse than `#[non_exhaustive]` as well.

And finally, being able to exhaustively match all errors, i.e. committing to exhaustiveness, is also a valuable service to library consumers.

RazrFalcon:
I personally hate `#[non_exhaustive]`, so I would rather have a _minor_ version bump.

adamreichold:
> I personally hate `#[non_exhaustive]`, so I would rather have a _minor_ version bump.

Fair enough.

RazrFalcon:
Done. Hopefully the changelog is good enough.

I genuinely do not remember most of the changes. It feels like someone's else codebase now. It barely has any code I wrote...
Still, thank you for all of your contributions. Someone had to work on them anyway. And we have a huge memory usage improvement in the end.

adamreichold:
> Done. Hopefully the changelog is good enough.

I don't think we did anything besides the namespace deduplication after v0.16.0. (I did some thinking-out-load-in-multiple-pull-requests-which-did-not-end-up-being-merged which might make it appear as if we did more changes.)

> I genuinely do not remember most of the changes. It feels like someone's else codebase now. It barely has any code I wrote...

I don't think that is [completely accurate](https://github.com/RazrFalcon/roxmltree/graphs/contributors). ;-) In any case, I'd say that multiple ongoing contributions is a positive sign of a healthy open source project. Thank you for writing and open sourcing roxmltree!

PR #89, opened 2023-01-15
Use consistent debug impls for the various iterators that avoid formatting the whole document.

### LINKED ISSUE #87: Less than usefull Debug output for Element
From roxmltree 0.15 to 0.16, the Debug output for `Element` changed from including the element name, attributes and namespaces to something similar but with an added virtual `doc` attribute including debug info for the entire document.  This changes diagnostic output in my program from a line of easy-to-digest text to thousands of not very comprehensible lines (my input xml in an example is 750 lines, which translates to 4559 lines (235 kB) out error message).

Is it possible to change this behaviour from code using roxmltree?  Would you accept a patch for skipping the `doc` virtual attribute in Debug formatting?

adamreichold:
`Descendants` actually had the same issue. I tried to resolve this consistently, i.e. I just left out the `doc` field everywhere and delegated to the existing iterator `Debug` impls (meaning they will print the iterator state, not a sequence of values). 

Closes #87

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -380,6 +380,7 @@ impl From<usize> for NodeId {
     }
 }
 
+#[derive(Debug)]
 enum NodeKind<'input> {
     Root,
     Element {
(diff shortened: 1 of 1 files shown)


RazrFalcon:
Thanks once again!

PR #92, opened 2023-01-22
Replace &'a str by &'a Cow<'input, str> where this is the backing data type.

### LINKED ISSUE #88: Most APIs rather return a `&'a str` than a `Cow<'input, str>`
I get that this is a bit more ergonomic, but it also means that users cannot do zero-copy stuff—or need to keep the `Document` around.
I think a solution could be to just make the fields that have Cows public and annotate that this is not a stable API.

adamreichold:
I started converting return types from `&'a str` to `&'a Cow<'input, str>` and there is indeed some loss ergonomics as calling code must call `.map(AsRef::as_ref)` to e.g. compare against `Option<&str>`.

We could work around this by providing `*_cow` methods returning `Option<&'a Cow<'input, str>>` and implement the existing methods as `.map(AsRef::as_ref)` on top of those so that no implementations need to be duplicated. But that would imply a significant increase in API surface for purely technical reasons.

Another issue that should however be solvable by adding another internal type is that `ExpandedName` does double duty a) holding references to expanded names from within the document to returned from `Node::tag_name` which would need to store `&'a Cow<'input, str>` and b) holding values passed into the API for matching purposes like `Node::has_tag_name` which need to store `&'c str` with an unrelated lifetime `'c`.

Closes #88

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -450,7 +450,7 @@ impl<'a, 'input> Attribute<'a, 'input> {
     /// assert_eq!(doc.root_element().attributes().nth(1).unwrap().name(), "a");
     /// ```
     #[inline]
-    pub fn name(&self) -> &'a str {
+    pub fn name(&self) -> &'input str {
         self.data.name.local_name
     }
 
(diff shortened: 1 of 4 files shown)


RazrFalcon:
I understand the performance benefits, but the new API is ugly as hell.
I haven't looked into it myself, so I'm not sure if there a better ways, but I would rather leave everything as is.

PR #93, opened 2023-01-23
Expose 'input lifetime where it is the backing lifetime.

adamreichold:
Broken out of #92 as it should be uncontroversial and that PR will most likely not land due to its ergonomic costs.


RazrFalcon:
Thanks.

benthillerkus:
Thank you!

PR #94, opened 2023-01-28
Use ref-counted owned strings.

RazrFalcon:
@adamreichold  how about this regarding #88 ?

Sure, it makes allocated strings a bit slower and introduces indirection. But it's the easiest way I can think of to preserve strings after dropping a `Document`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,7 +31,7 @@ use core::fmt;
 use core::hash::{Hash, Hasher};
 use core::num::NonZeroU32;
 
-use alloc::rc::Rc;
+use alloc::sync::Arc;
 use alloc::vec::Vec;
 
 pub use xmlparser::TextPos;
@@ -411,7 +411,7 @@ pub enum SharedString<'input> {
     /// A raw slice of the input string.
     Borrowed(&'input str),
     /// A reference-counted String.
-    Owned(Rc<str>),
+    Owned(Arc<str>),
 }
 
 impl SharedString<'_> {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -385,7 +385,7 @@ enum NodeKind<'input> {
         namespaces: ShortRange,
     },
     PI(PI<'input>),
-    Comment(&'input str),
+    Comment(StringStorage<'input>),
     Text(StringStorage<'input>),
 }
 
@@ -1135,6 +1135,7 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
     /// Returns node's text.
     ///
     /// - for an element will return a first text child
+    /// - for a comment will return a self text
     /// - for a text node will return a self text
     ///
     /// # Examples
@@ -1150,6 +1151,12 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
     ///            Some("\n    text\n"));
     /// assert_eq!(doc.root_element().first_child().unwrap().text(),
     ///            Some("\n    text\n"));
+    /// ```
+    ///
+    /// ```
+    /// let doc = roxmltree::Document::parse("<!-- comment --><e/>").unwrap();
+    ///
+    /// assert_eq!(doc.root().first_child().unwrap().text(), Some(" comment "));
     /// ```
     #[inline]
     pub fn text(&self) -> Option<&'a str> {
@@ -1168,6 +1175,7 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
                 },
                 _ => None,
             },
+            NodeKind::Comment(ref text) => Some(text),
             NodeKind::Text(ref text) => Some(text),
             _ => None,
         }
@@ -1208,23 +1216,6 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
                 _ => None,
             },
             None => None,
-        }
-    }
-
-    /// Returns node's comment content.
-    ///
-    /// # Examples
-    ///
-    /// ```
-    /// let doc = roxmltree::Document::parse("<!-- comment --><e/>").unwrap();
-    ///
-    /// assert_eq!(doc.root().first_child().unwrap().comment(), Some(" comment "));
-    /// ```
-    #[inline]
-    pub fn comment(&self) -> Option<&'input str> {
-        match self.d.kind {
-            NodeKind::Comment(text) => Some(text),
-            _ => None,
         }
     }
 
@@ -1409,8 +1400,8 @@ impl<'a, 'input: 'a> fmt::Debug for Node<'a, 'input> {
             NodeKind::PI(pi) => {
                 write!(f, "PI {{ target: {:?}, value: {:?} }}", pi.target, pi.value)
             }
-            NodeKind::Comment(text) => write!(f, "Comment({:?})", text),
-            NodeKind::Text(ref text) => write!(f, "Text({:?})", text),
+            NodeKind::Comment(ref text) => write!(f, "Comment({:?})", text.as_str()),
+            NodeKind::Text(ref text) => write!(f, "Text({:?})", text.as_str()),
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -449,7 +449,7 @@ impl StringStorage<'_> {
     /// Returns a string slice.
     pub fn as_str(&self) -> &str {
         match self {
-            StringStorage::Borrowed(ref s) => s,
+            StringStorage::Borrowed(s) => s,
             StringStorage::Owned(s) => s,
         }
     }
@@ -457,7 +457,7 @@ impl StringStorage<'_> {
 
 impl PartialEq for StringStorage<'_> {
     fn eq(&self, other: &Self) -> bool {
-        &*self == &*other
+        self.as_str() == other.as_str()
     }
 }
 
@@ -542,13 +542,12 @@ impl<'a, 'input> Attribute<'a, 'input> {
         &self.data.value
     }
 
-    /// Returns attribute's shared value.
+    /// Returns attribute's value storage.
     ///
-    /// Unlike a string returned by `value()`, `StringStorage` can outlive the [`Document`].
-    #[cfg(any(feature = "rc-strings", feature = "arc-strings"))]
+    /// Useful when you need a more low-level access to an allocated string.
     #[inline]
-    pub fn shared_value(&self) -> StringStorage<'input> {
-        self.data.value.clone()
+    pub fn value_storage(&self) -> &StringStorage<'input> {
+        &self.data.value
     }
 
     /// Returns attribute's position in bytes in the original document.
@@ -1136,7 +1135,6 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
     /// Returns node's text.
     ///
     /// - for an element will return a first text child
-    /// - for a comment will return a self text
     /// - for a text node will return a self text
     ///
     /// # Examples
@@ -1153,14 +1151,15 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
     /// assert_eq!(doc.root_element().first_child().unwrap().text(),
     ///            Some("\n    text\n"));
     /// ```
-    ///
-    /// ```
-    /// let doc = roxmltree::Document::parse("<!-- comment --><e/>").unwrap();
-    ///
-    /// assert_eq!(doc.root().first_child().unwrap().text(), Some(" comment "));
-    /// ```
     #[inline]
     pub fn text(&self) -> Option<&'a str> {
+        self.text_storage().map(|s| s.as_str())
+    }
+
+    /// Returns node's text storage.
+    ///
+    /// Useful when you need a more low-level access to an allocated string.
+    pub fn text_storage(&self) -> Option<&'a StringStorage<'input>> {
         match self.d.kind {
             NodeKind::Element { .. } => match self.first_child() {
                 Some(child) if child.is_text() => match self.doc.nodes[child.id.get_usize()].kind {
@@ -1169,7 +1168,6 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
                 },
                 _ => None,
             },
-            NodeKind::Comment(text) => Some(text),
             NodeKind::Text(ref text) => Some(text),
             _ => None,
         }
@@ -1193,6 +1191,13 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
     /// ```
     #[inline]
     pub fn tail(&self) -> Option<&'a str> {
+        self.tail_storage().map(|s| s.as_str())
+    }
+
+    /// Returns element's tail text storage.
+    ///
+    /// Useful when you need a more low-level access to an allocated string.
+    pub fn tail_storage(&self) -> Option<&'a StringStorage<'input>> {
         if !self.is_element() {
             return None;
         }
@@ -1203,6 +1208,23 @@ impl<'a, 'input: 'a> Node<'a, 'input> {
                 _ => None,
             },
             None => None,
+        }
+    }
+
+    /// Returns node's comment content.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// let doc = roxmltree::Document::parse("<!-- comment --><e/>").unwrap();
+    ///
+    /// assert_eq!(doc.root().first_child().unwrap().comment(), Some(" comment "));
+    /// ```
+    #[inline]
+    pub fn comment(&self) -> Option<&'input str> {
+        match self.d.kind {
+            NodeKind::Comment(text) => Some(text),
+            _ => None,
         }
     }
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,7 +31,6 @@ use core::fmt;
 use core::hash::{Hash, Hasher};
 use core::num::NonZeroU32;
 
-use alloc::sync::Arc;
 use alloc::vec::Vec;
 
 pub use xmlparser::TextPos;
@@ -310,7 +309,6 @@ struct ShortRange {
 impl From<Range> for ShortRange {
     #[inline]
     fn from(range: Range) -> Self {
-        // Casting to `u32` should be safe since we have a 4GiB input data limit.
         debug_assert!(range.start <= core::u32::MAX as usize);
         debug_assert!(range.end <= core::u32::MAX as usize);
         ShortRange::new(range.start as u32, range.end as u32)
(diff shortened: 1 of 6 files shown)


adamreichold:
I think the discussion from #71 applies, i.e. this is significantly hurting the case where I do not want to take ownership of the strings by adding reference counting and it suddenly makes `Document: !Send` which would for example significantly hurt ergonomics for myself using roxmltree in multi-threaded async server code. The alternative of using `Arc<str>` would hurt the single-threaded case and code not interested in taking ownership even more.

So in summary, I think shared ownership via reference counting moves too far away from roxmltree's original design goals. And while it will be much harder to implement, I think providing on-demand normalization which would return `Cow<'input, str>` by value anyway would be preferable to the approach presented here.

As a technicality, we should probably use `Rc<str>` instead of `Rc<String>` since long-term the shared text is immutable anyway, but of course this prevents us from using `&mut String` via `Rc::make_mut` while we build up the text values. (But that part of the parser is a bit of a code smell anyway IMHO as it is the one place we were modify (instead of just building) the read-only tree.) ((As an aside, we have a `&mut SharedString` and hence `&mut Rc<String>` there, so it should be possible to use `Rc::get_mut` instead of `Rc::make_mut`.))

RazrFalcon:
Are you sure about the performance? I just checked and it's identical. After all, we're rarely allocating strings.
`Arc` has ~1% performance penalty, but it's neglectable. 

Overall, I agree that this a complicated design topic. I have a use case (`resvg`), where currently I allocate almost all string anyway. Because dealing with lifetimes becomes too annoying/impossible.
But even in this case using shared strings will improve the performance, since large attributes almost always have new lines, i.e. normalized, and I'm needlessly reallocating them.

I've tried storing `SharedString` directly in my own data structure, but lifetimes are a mess. In the end, I wasn't able to compile my code. So while it would be the perfect solution, the implementation would require some extreme lifetimes gymnastics. Afaik I would need three lifetimes to achieve what I want.

And as you're already mentioned, delayed normalization would require preserving too much info, making everything way too complicated.

Anyway, I'm fine with `Arc`, if this is satisfies your use case. I think this change is a good middle-ground.

adamreichold:
> Are you sure about the performance? I just checked and it's identical. After all, we're rarely allocating strings.
Arc has ~1% performance penalty, but it's neglectable.

Not without measuring. Just two caveats I see:
* This does depend on the input documents and if the argument is that we seldom need to allocate strings, then cloning an exposed `&'a Cow<'input, str>` falls under the same argument without the costs of reference counting.
* The cache coherence traffic introduced by using atomics is harder to measure in isolated benchmarks as it measuring it depends on produce load on multiple cores.

> Overall, I agree that this a complicated design topic. I have a use case (resvg), where currently I allocate almost all string anyway. Because dealing with lifetimes becomes too annoying/impossible.
But even in this case using shared strings will improve the performance, since large attributes almost always have new lines, i.e. normalized, and I'm needlessly reallocating them.

You mean that you end up storing `Arc<str>` in your structures, i.e. bumping the reference count for normalized attribute values and allocating/copying for unnormalized attribute values? Doesn't this pessimize the unnormalized attribute values? What if they are much more common?

src/lib.rs
```diff
@@ -402,10 +403,51 @@ struct NodeData<'input> {
...
+    Borrowed(&'input str),
+    /// A reference-counted String.
+    Owned(Arc<str>),
+}
+
+impl SharedString<'_> {
+    /// Returns a string slice.
+    pub fn as_str(&self) -> &str {
+        match self {
+            SharedString::Borrowed(ref s) => s,
```
adamreichold, line 421:
```suggestion
            SharedString::Borrowed(s) => s,
```

src/lib.rs
```diff
@@ -402,10 +403,51 @@ struct NodeData<'input> {
...
+            SharedString::Owned(s) => s,
+        }
+    }
+}
+
+impl PartialEq for SharedString<'_> {
+    fn eq(&self, other: &Self) -> bool {
+        &*self == &*other
```
adamreichold, line 429:
```suggestion
        self.as_str() == other.as_str()
```

src/lib.rs
```diff
@@ -469,6 +511,14 @@ impl<'a, 'input> Attribute<'a, 'input> {
         &self.data.value
     }
 
+    /// Returns attribute's shared value.
+    ///
+    /// Unlike a string returned by `value()`, `SharedString` can outlive the [`Document`].
+    #[inline]
+    pub fn shared_value(&self) -> SharedString<'input> {
```
adamreichold, line 518: I think it would be more flexible to expose `&SharedString<'input>` and let the caller clone if necessary.

src/parse.rs
```diff
@@ -1028,14 +1028,18 @@ fn append_text<'input, 'temp>(
...
                         let mut concat_text = String::with_capacity(s.len() + text.len());
                         concat_text.push_str(s);
                         concat_text.push_str(text);
-                        *prev_text = Cow::Owned(concat_text);
+                        *prev_text = SharedString::Owned(Arc::from(concat_text));
                     }
-                    Cow::Owned(ref mut s) => {
-                        s.push_str(text);
+                    SharedString::Owned(s) => {
+                        // TODO: find a way to not reallocate the string.
```
adamreichold, line 1038: I am really weary about this part, because with `Arc<str>` it means dropping excess capacity when concatenating the strings. I wonder if a type like

```rust
pub enum Text<'input> {
  Borrowed(&'input str),
  Owned(String),
  Shared(Arc<str>),
}
```

would be nicer because of two reasons:
* We could use `Text::Owned(..)` while extending text nodes like here.
* Going from `Text::Owned(..)` to `Text::Shared(..)` could be single optional post-processing step when we are sure to never touch a text node again.

If that post-processing step to go from `Text::Owned(..)` to `Text::Shared(..)` would be controlled by `ParseOptions`, this would also mean that users who do not want shared text could avoid the reference counting completely. (We already pay for the enum tag with `Cow`. Sadly, we cannot make `Arc` versus `Rc` decidable at runtime.)

adamreichold: Actually, the sharing could be fully lazy without an option in `ParseOptions`: We always produce `Text::Owned(..)` during parse and your `shared_value` becomes:

```
pub fn shared_value(&mut self) -> Arc<str> {
  match self.data.value {
    Text::Borrowed(text) => {
      let text = Arc::from(text),
      self.data.value = Text::Shared(text.clone());
      text      
    },
    Text::Owned(text) => {
      let text = Arc::from(text);
      self.data.value = Text::Shared(text.clone());
      text
    }
    Text::Shared(text) => text.clone(),
  }
}
```

adamreichold: That way, we do not pay anything we are not already paying and `shared_value` et al. are also not paying anything they do not in the up-front approach expect for the `&mut self` requirement.

RazrFalcon:
>The cache coherence traffic introduced by using atomics is harder to measure in isolated benchmarks as it measuring it depends on produce load on multiple cores.

I understand that `Arc` isn't free, but I cannot imagine a use case where it would become a bottleneck. Especially compared with XML parsing itself.
You're optimizing for an unrealistic case.

>then cloning an exposed &'a Cow<'input, str> falls under the same argument without the costs of reference counting

But it will reallocate a string, while Arc is just a ref bump.

> You mean that you end up storing Arc<str> in your structures, i.e. bumping the reference count for normalized attribute values and allocating/copying for unnormalized attribute values? 

Yes. Before I was allocating all strings, so it couldn't get any worse than that. Sure, it would be great to keep original `&str`, but I couldn't solve myriad of lifetime issues.
And even if I would be able to solve lifetime issues, I would still benefit from `Arc` and not `Cow`.

Honestly, it seems like the easier solution would be to simply have:

```rust
enum SharedString {
    Borrowed(Range<usize>),
    Owned(Arc<str>),
}
```

This way we can forget about lifetimes completely :/

adamreichold:
> You're optimizing for an unrealistic case.

As discussed above, I think with on-demand sharing, we can have our cake and it too. When `Arc<str>` is created only for those strings on which `shared_value` is actually called, I would have absolutely no objections.

I do think `fn as_shared(&mut self) -> Arc<str>` should become a method on `Text` tough as we'll need for more than just attribute values. I do not think that `enum Text` needs to be public though as we never expose it in the API if we have only methods like `shared_value(&mut self) -> Arc<str>`.

src/lib.rs
adamreichold: Not applicable if we go for on-demand sharing as detailed below.

src/parse.rs
adamreichold: So this "wonderful" idea of on-demand sharing does not work as `Attribute` has not access to `&mut AttributeData`. The `ParserOption` route might still be viable though. I am currently looking into that...

adamreichold: I tried to resolve this TODO but I do not think it can be done safely as going from `String` to `Arc<str>` will copy the text into a new allocation where it is prepend with `Arc`'s header. It is then not possible to get a mutable variant like `String` back for this allocation without doing something about that header, e.g. dropping it thereby requiring a new allocation. (This also means that in the case of the previous node being borrowed, we also do one more copy of the concatenated text. It would surprise me if benchmarks cannot detect this amount of additional copying?)

src/lib.rs
adamreichold: But probably a requirement if this implemented for e.g. text nodes as we other need to duplicate the implementation of `fn text(&self) -> ...`. With `fn shared_text(&self) -> &SharedString<'input>` we could implement `fn text(&self) -> &str` in terms of that.

src/parse.rs
adamreichold: So since this is copying anyway, we only need to amortize the extra allocations, e.g.

```diff
diff --git a/src/parse.rs b/src/parse.rs
index 36cb3ab..a4e9475 100644
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -370,6 +370,7 @@ struct ParserData<'input> {
     parent_prefixes: Vec<&'input str>,
     entities: Vec<Entity<'input>>,
     buffer: TextBuffer,
+    concat_buffer: String,
     after_text: bool,
 }
 
@@ -481,6 +482,7 @@ fn parse(text: &str, opt: ParsingOptions) -> Result<Document, Error> {
         awaiting_subtree: Vec::new(),
         parent_prefixes: Vec::new(),
         buffer: TextBuffer::new(),
+        concat_buffer: String::new(),
         after_text: false,
     };
 
@@ -585,6 +587,7 @@ fn process_tokens<'input>(
                     parent_id,
                     span.start(),
                     pd.opt.nodes_limit,
+                    &mut pd.concat_buffer,
                     pd.after_text,
                     doc,
                     &mut pd.awaiting_subtree,
@@ -910,6 +913,7 @@ fn process_text<'input>(
             parent_id,
             text.start(),
             pd.opt.nodes_limit,
+            &mut pd.concat_buffer,
             pd.after_text,
             doc,
             &mut pd.awaiting_subtree,
@@ -953,6 +957,7 @@ fn process_text<'input>(
                         parent_id,
                         text.start(),
                         pd.opt.nodes_limit,
+                        &mut pd.concat_buffer,
                         pd.after_text,
                         doc,
                         &mut pd.awaiting_subtree,
@@ -980,6 +985,7 @@ fn process_text<'input>(
             parent_id,
             text.start(),
             pd.opt.nodes_limit,
+            &mut pd.concat_buffer,
             pd.after_text,
             doc,
             &mut pd.awaiting_subtree,
@@ -1012,12 +1018,13 @@ impl<'input, 'temp> BorrowedText<'input, 'temp> {
     }
 }
 
-#[allow(clippy::needless_lifetimes)]
+#[allow(clippy::needless_lifetimes, clippy::too_many_arguments)]
 fn append_text<'input, 'temp>(
     text: BorrowedText<'input, 'temp>,
     parent_id: NodeId,
     pos: usize,
     nodes_limit: u32,
+    concat_buffer: &mut String,
     after_text: bool,
     doc: &mut Document<'input>,
     awaiting_subtree: &mut Vec<NodeId>,
@@ -1027,21 +1034,12 @@ fn append_text<'input, 'temp>(
         if let Some(node) = doc.nodes.last_mut() {
             if let NodeKind::Text(ref mut prev_text) = node.kind {
                 let text = text.as_str();
-                match prev_text {
-                    SharedString::Borrowed(s) => {
-                        let mut concat_text = String::with_capacity(s.len() + text.len());
-                        concat_text.push_str(s);
-                        concat_text.push_str(text);
-                        *prev_text = SharedString::Owned(Arc::from(concat_text));
-                    }
-                    SharedString::Owned(s) => {
-                        // TODO: find a way to not reallocate the string.
-                        let mut concat_text = String::with_capacity(s.len() + text.len());
-                        concat_text.push_str(s);
-                        concat_text.push_str(text);
-                        *prev_text = SharedString::Owned(Arc::from(concat_text));
-                    }
-                }
+                let s = prev_text.as_str();
+                concat_buffer.clear();
+                concat_buffer.reserve(s.len() + text.len());
+                concat_buffer.push_str(s);
+                concat_buffer.push_str(text);
+                *prev_text = SharedString::Owned(Arc::from(concat_buffer.as_str()));
             }
         }
     } else {
```

adamreichold:
This still uses `into_cow` for going from `BorrowedText` to `SharedString` which does not really fit. I would suggest using `Text` instead of `SharedString` and thereby `into_text` for the method. Making it an implementation detail and putting it next to `BorrowedText` would seem helpful to me as well.

RazrFalcon:
I will put in on hold until the next weekend.

I still do not see how `Arc` can become a bottleneck, since 99% of the time would be spent on XML parsing alone. And attribute value normalization is triggered pretty rarely.

Anyway, in my use case I still cannot store `SharedString` directly, because of lifetime issues. And unless I will be able to solve them I would not need this change anyway.

I guess returning just a `&str` is our only option. Otherwise it's either `Arc` or mutable tree.

RazrFalcon:
Ok, I was able to beat lifetimes. Now this patch works perfectly for my use case.
So I will probably merge it next week.

`Cow` is no go for me, because cloning would be way too expensive. 
This leaves us with normalization on access, by returning `Cow<str>` instead of `&str`, which is meh.
Or something else...

Also, I think it would be a good idea to make Arc optional, via a build feature. To switch between `Rc` and `Arc`.

adamreichold:
I am resigned to `Arc<str>` now personally, the only open discussions I see here are comments on the current implementation. I would be happy if they could be addressed before merging this.

> Also, I think it would be a good idea to make Arc optional, via a build feature. To switch between Rc and Arc.

If we do that, we could theoretically have two features, e.g. `rc` for non-thread-safe shared text and `arc` for thread-safe shared text and just `Owned(String)` if neither is enabled.

RazrFalcon:
Yes, I will address your comments.

Yes, I guess we can do this. A better solution would be to have `Document` generic over the string storage type, but it would  make code even more complex.

src/parse.rs
RazrFalcon: I think this can be done separately. Afaik, the second branch is executed only when we have sequential CDATA, which is extremely rare.

src/lib.rs
RazrFalcon: As always, it wasn't as easy. We store comment as a string and return it via `Node::text`. Meaning we cannot return `&SharedString` here.

We should either store comment as `SharedString` as well or have a separate method for getting the comment. Which is what I did. This is a breaking change, but a minor one.

RazrFalcon:
I think I have addressed everything. I do not plan merging it until the weekend anyway.

src/parse.rs
adamreichold: Agreed. I volunteer for a follow-up PR. :-) (The buffer would also amortize the allocation in the first branch.)

src/lib.rs
```diff
@@ -402,10 +399,86 @@ struct NodeData<'input> {
     pos: usize,
 }
 
+#[cfg(all(feature = "rc-strings", feature = "arc-strings"))]
+compile_error!("only one shared string type can be enable");
```
adamreichold, line 403: This might break builds if two crates in the dependency tree enable both features. Which might be unfixable at the top level. I would suggest to make this "more additive" by giving one feature precedence. I think `Arc` over `Rc` would make sense as it has strictly more functionality similar to both having more functionality than `String`?

adamreichold (approved):
LGTM. The only thing I would suggest is to make the feature additive. (You left out storage access to namespace URI, but that does admittedly seem somewhat niche.)

src/lib.rs
RazrFalcon: Hmm... maybe. I'm not very familiar with cargo's features resolving logic. I would thought that it would simply instance the library twice.

But automatically switching to Arc isn't the best idea either. What if someones code strictly depends on it being `Rc`? Then it would not compile either.

Seems like having a generic Document is the perfect solution, but we already have two lifetimes. Adding a generic type to all types would be a nightmare...

adamreichold: I think I would prefer the possible breakage of exclusive features to the hassle of a type parameter. 

The `im` crate which faces the same problem went for a completely separate crate to avoid the above, c.f. https://docs.rs/im/latest/im/#thread-safety, which would seem a bit excessive for our use case though IMHO.

> What if someones code strictly depends on it being Rc? Then it would not compile either.

I think this applies just as well to `String` versus `Rc` as it does to `Rc` versus `Arc`, doesn't it?

RazrFalcon: Hmm... I guess having multiple storage types isn't as easy as I thought. A type parameter is the only "correct" option. Will think about it more.

RazrFalcon: In the end, this is basically Arc-only or else...

adamreichold: I only now realized that with the current implementation, every variant of the library pays the same reallocation and copying costs for string concatenation. So it is probably fine to go `Arc`-only and avoid all this trouble.

RazrFalcon: Yes, the main cost is atomic refs in case of Arc and indirection(?) in case of Rc/Arc.

But looks like the only way to have custom string storage is a type parameter, which is no go. 
I will update the patch to Arc.

src/lib.rs
```diff
@@ -402,10 +399,62 @@ struct NodeData<'input> {
...
+    /// A raw slice of the input string.
+    Borrowed(&'input str),
+
+    /// A reference-counted string.
+    Owned(alloc::sync::Arc<str>),
+}
+
+impl StringStorage<'_> {
+    /// Creates a new owned string from `&str` or `String`.
+    pub fn new_owned<T: Into<alloc::sync::Arc<str>>>(s: T) -> Self {
```
adamreichold, line 421: Does this method really carry its weight now? The enum variant is public in any case and the trait bounds are the same as when constructing the enum at the single place where this is used?

RazrFalcon: The difference is that when I have to create my own `StringStorage`, which I do, it will be:

```rust
StringStorage::new_owned("text")
// instead of
StringStorage::Owned(Arc::from("text"))
```

A nice shorthand.

PR #95, opened 2023-01-28
Runtime-controlled sharing of attribute values

adamreichold:
This is a small modification of #94 which has the same API but a more complicated operation, i.e. for `shared_value` to be efficient (a reference count bump), `ParsingOptions::shared_text` must be set, otherwise this is just copying the string as with `Cow`.

To be honest, the above is complicated will only be noticed due to performance regressions. An alternative would be `fn shared_value(&self) -> Option<&Arc<str>>` which will not copy as a fallback, but that still leaves the need to set `shared_text` and `.unwrap()` the `Option`.

So maybe this is not worth avoiding the small `Arc`-related cost. I am not really happy either way.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -26,7 +26,6 @@ extern crate alloc;
 #[cfg(feature = "std")]
 extern crate std;
 
-use alloc::borrow::Cow;
 use alloc::vec::Vec;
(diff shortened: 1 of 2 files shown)


adamreichold:
Closing this as the API is too complicated.

Issue #108, opened 2023-10-16
Encoding gets messed up for CDATA


WhyNotHugo:
This applies to master and not to 0.18.1.

Copy this into `src/lib.rs` and run `cargo test`:

```rs
#[test]
fn test_multi_get_parse_encoding_another() {
    let calendar_data = ExpandedName::from(("urn:ietf:params:xml:ns:caldav", "calendar-data"));

    let body = "<?xml version=\"1.0\" encoding=\"utf-8\"?>\n<C:calendar-data xmlns:C=\"urn:ietf:params:xml:ns:caldav\"><![CDATA[BEGIN(de baño)VCALENDAR\r\n]]></C:calendar-data>\n";
    assert!(body.contains("baño"));

    let body = std::str::from_utf8(body.as_ref()).unwrap();
    assert!(body.contains("baño"));

    let doc = Document::parse(body).unwrap();

    let raw_data = doc
        .descendants()
        .find(|node| node.tag_name() == calendar_data)
        .unwrap()
        .text()
        .unwrap();
    std::dbg!(&raw_data);
    assert!(raw_data.contains("baño"));
}
```

Basically this parses a string to a `roxmltree` and then reads the text from the cdata. `baño` turns into `baÃ±o` (that's why the `std::dbg!` is there).

WhyNotHugo:
This regresses in 4f3566bcb919f63af6ad57e69bcbf7e03236968b.

WhyNotHugo:
The issue is in `StringExt`. The way it pushes `u8` into `String` is not correct. Consider:

```rs
fn main() {
    let mut s = String::new();
    for c in String::from("baño").bytes() {
        s.push(c as char);
    }

    assert_eq!("baÃ±o", s);
}
````

WhyNotHugo:
Converting individual bytes to characters and pushing those individually is not the same as pushing a single multi-byte character.

If I understand correctly, converting `u8` to `char` uses ISO-8859-1: https://doc.rust-lang.org/std/primitive.char.html#impl-From%3Cu8%3E-for-char

RazrFalcon:
Ugh... my bad. Didn't thought about that.

PR #109, opened 2023-10-16
Don't split characters into bytes for processing

WhyNotHugo:
Fixes: https://github.com/RazrFalcon/roxmltree/issues/108

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -1277,19 +1277,19 @@ impl StringExt for String {
     //
     // https://www.w3.org/TR/xml/#sec-line-ends
     fn push_from_text(&mut self, c: char, at_end: bool) {
-        if self.as_bytes().last() == Some(&b'\r') {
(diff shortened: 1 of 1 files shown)


WhyNotHugo:
I can copy the test from https://github.com/RazrFalcon/roxmltree/issues/108, I'm just not sure where is the right place.

WhyNotHugo:
This fixes the related issue that affects me, but I think that in general, the input shouldn't be treated as bytes but as chars, given that it's already validated utf-8.

RazrFalcon:
I don't think this is the correct solution. You handle only the text node case. I will partially revert the old one.

WhyNotHugo:
Yeah, I think that this is a partial solution (for my specific case), but I’m not entirely sure how to handle other cases or what those might be.

Didn’t the original commit fix a bug with \r\n ?

RazrFalcon:
No. It was separate. And it would not be affected.

PR #120, opened 2024-05-22
Add `range_qname` and `range_value` methods to `Attribute`.

Jayonas:
To save memory this only uses a `u16` and `u8` per `Attribute`, but the tradeoff is that the qname and value ranges will be incorrect in a few extreme data cases, specifically if the attribute's qname has len > 65535 or there are >254 spaces surrounding its equal sign.  Those cases seem unlikely enough in practice to justify not using more memory for this.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -490,9 +490,9 @@ struct AttributeData<'input> {
     value: StringStorage<'input>,
     #[cfg(feature = "positions")]
     range: Range<usize>,
-    #[cfg(feature = "positions-extra-attr")]
-    qname_len: u8,
-    #[cfg(feature = "positions-extra-attr")]
+    #[cfg(feature = "positions")]
+    qname_len: u16,
+    #[cfg(feature = "positions")]
     eq_len: u8, // includes any surrounding spaces
 }
 
@@ -599,12 +599,12 @@ impl<'a, 'input> Attribute<'a, 'input> {
     ///    ^^^^^^
     /// ```
     ///
-    /// This method will return incorrect data if the attribute's qname is longer than 255 bytes,
-    /// although it will not panic or exhibit undefined behavior.
-    #[cfg(feature = "positions-extra-attr")]
+    /// To reduce memory usage the qname length is limited by u16::MAX.
+    /// If the attribute exceeds that limit then the end of the returned range will be incorrect.
+    #[cfg(feature = "positions")]
     #[inline]
     pub fn range_qname(&self) -> Range<usize> {
-        let end = self.data.range.start + self.data.qname_len as usize;
+        let end = self.data.range.start + usize::from(self.data.qname_len);
         self.data.range.start..end
     }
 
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -587,6 +591,43 @@ impl<'a, 'input> Attribute<'a, 'input> {
...
+    /// <e n:attr='value'/>
+    ///    ^^^^^^
+    /// ```
+    ///
+    /// This method will return incorrect data if the attribute's qname is longer than 255 bytes,
+    /// although it will not panic or exhibit undefined behavior.
+    #[cfg(feature = "positions-extra-attr")]
+    #[inline]
+    pub fn range_qname(&self) -> Range<usize> {
+        let end = self.data.range.start + self.data.qname_len as usize;
```
RazrFalcon, line 607: I would prefer `usize::from(self.data.qname_len)` here.

src/lib.rs
```diff
@@ -490,6 +490,10 @@ struct AttributeData<'input> {
     value: StringStorage<'input>,
     #[cfg(feature = "positions")]
     range: Range<usize>,
+    #[cfg(feature = "positions-extra-attr")]
+    qname_len: u8,
```
RazrFalcon, line 494: We can easily use `u16` here. the struct would be padded to 64B anyway. We have plenty of space.

src/lib.rs
```diff
@@ -587,6 +591,43 @@ impl<'a, 'input> Attribute<'a, 'input> {
...
     }
+
+    /// Returns attribute's qname's range in bytes in the original document.
+    ///
+    /// ```text
+    /// <e n:attr='value'/>
+    ///    ^^^^^^
+    /// ```
+    ///
+    /// This method will return incorrect data if the attribute's qname is longer than 255 bytes,
```
RazrFalcon, line 602: I would rather say:

To reduce memory usage the qname length is limited by `u16::MAX`.

No need to mention panics or undefined behavior.

src/lib.rs
```diff
@@ -587,6 +591,43 @@ impl<'a, 'input> Attribute<'a, 'input> {
...
+    /// Returns attribute's value's range in bytes in the original document, excluding the surrounding quotes.
+    ///
+    /// If the attribute's value is an empty string then the `start` and `end` of this `Range` are equal, and indicate the closing quote.
+    ///
+    /// ```text
+    /// <e n:attr='value'/>
+    ///            ^^^^^
+    /// ```
+    ///
+    /// This method will return incorrect data if the attribute's qname is longer than 255 bytes
```
RazrFalcon, line 620: Same here.

tests/api.rs
```diff
@@ -202,6 +216,79 @@ fn text_pos_03() {
...
+        assert_eq!(doc.text_pos_at(attr.range().start), TextPos::new(1, 36));
+        assert_eq!(doc.text_pos_at(attr.range().end), TextPos::new(1, 48));
+        assert_eq!(doc.text_pos_at(attr.range_qname().start), TextPos::new(1, 36));
+        assert_eq!(doc.text_pos_at(attr.range_qname().end), TextPos::new(1, 40));
+        assert_eq!(doc.text_pos_at(attr.range_value().start), TextPos::new(1, 47));
+        assert_eq!(doc.text_pos_at(attr.range_value().end), TextPos::new(1, 48));
+    }
+}
+
+#[cfg(feature = "positions-extra-attr")]
```
RazrFalcon, line 255: It's fine to skip those this test. u16 is plenty enough.

src/lib.rs
Jayonas: Done, but I also added a sentence that concretely explains what effect this limit has on the returned value because I don't think it's necessarily obvious (it probably is for `range_qname`, but not for `range_value`).  Happy to change or remove that if you'd like.

tests/api.rs
Jayonas: Wasn't sure if you meant to remove both tests, or only the one for qname exceeding the limit (I agree that one's unnecessary if the limit is `u16`).  For now I left the test for `eq_len` maxing out, but I can remove that one too if you want.

RazrFalcon:
Thanks! Looks good to me. Is that all or do you plan any other changes? If no then I will publish a new release.

Jayonas:
Nope, not planning any other changes right now.  Thanks!

PR #121, opened 2024-05-26
RFC: Add note on availability of serde-roxmltree as an additional API layer.

adamreichold:
We have been happily using `serde-roxmltree` in addition to the native API for our XML parsing needs for quite some time now and I have recently added opt-in support for namespaces, so by now I feel like `serde-roxmltree` can be a useful way to reduce the boiler plate of the iterator-based API when deserializing simple documents.

Hence I think it would be helpful to make potential users of this crate aware of the "add-on" crate. Of course, I understand if you want to avoid playing favourites or being associated with a crate which you do not control, hence the RFC moniker.


RazrFalcon:
Sure. Sounds good.

PR #130, opened 2025-01-12
Add ASCII-only fast path also for checking NameStartChar in qualified names.

adamreichold:
Benchmarks are somewhat noisy on this one even though I think this should match the distribution of qualified names in the wild...

### DIFF
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -25,7 +25,7 @@ impl XmlCharExt for char {
     fn is_xml_name_start(&self) -> bool {
         // Check for ASCII first.
         if *self as u32 <= 128 {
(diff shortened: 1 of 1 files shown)


RazrFalcon:
Not sure about this one. It make code much more complicated without much performance improvement. 

We might even want to add a function that accepts the whole `&str`, to reduce code duplication.
Having `if (b < 128 && !b.is_xml_name_start()) || (b >= 128 && check_unicode_xml_name_start(prefix)) {` twice is a bad idea.

adamreichold:
> Not sure about this one. It makes code much more complicated without much performance improvement.

I agree. This was something that looked good on paper but does not appear to meaningfully improve things. Will adjust it to only add the ASCII fast path for the name start check.

PR #131, opened 2025-01-12
Devirtualize XmlEvents dispatch to give the optimizer full visibility

adamreichold:
As `XmlEvents` is not public API any more, there should be no possibility of code bloat due to multiple instantiations and we can safely monomorphise dispatch from the tokenizer into the parser.

This yields small but measurable improvements in the benchmarks, but only after marking the single implementation of `XmlEvents::token` as `#[inline(always)]`, so possibly by avoiding one level of dispatching on the token type, c.f. before

```
test huge_roxmltree                         ... bench:   5,394,963 ns/iter (+/- 143,581)
test large_roxmltree                        ... bench:   2,646,532 ns/iter (+/- 19,663)
test medium_roxmltree                       ... bench:     698,086 ns/iter (+/- 70,686)
test tiny_roxmltree                         ... bench:       4,236 ns/iter (+/- 99)
```

and after

```
test huge_roxmltree                         ... bench:   5,167,046 ns/iter (+/- 227,797)
test large_roxmltree                        ... bench:   2,422,624 ns/iter (+/- 47,267)
test medium_roxmltree                       ... bench:     685,669 ns/iter (+/- 23,478)
test tiny_roxmltree                         ... bench:       3,936 ns/iter (+/- 56)
```

### DIFF
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -608,6 +608,7 @@ fn parse(text: &str, opt: ParsingOptions) -> Result<Document> {
 }
 
 impl<'input> tokenizer::XmlEvents<'input> for Context<'input> {
+    #[inline(always)]
     fn token(&mut self, token: tokenizer::Token<'input>) -> Result<()> {
         match token {
             tokenizer::Token::ProcessingInstruction(target, value, range) => {
--- a/src/tokenizer.rs
+++ b/src/tokenizer.rs
@@ -192,7 +192,7 @@ pub trait XmlEvents<'input> {
 pub fn parse<'input>(
     text: &'input str,
     allow_dtd: bool,
-    events: &mut dyn XmlEvents<'input>,
+    events: &mut impl XmlEvents<'input>,
 ) -> Result<()> {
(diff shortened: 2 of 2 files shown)


adamreichold:
> I honestly would suggest passing `Context` directly. `XmlEvents` is just an artifact of old codebase.

I did that but the tokenizer tests still use the interface to collect raw streams of tokens. While it is an artifacts, this way of doing it allows us to reap the performance benefits without having to apply conditional compilation or some such for the tokenizer tests.

RazrFalcon:
Passing `Context` directly would also make code a bit cleaner, but sure, we can do it later.

adamreichold:
> Passing Context directly would also make code a bit cleaner, but sure, we can do it later.

It would allow to completely drop `Context::token` and theoretically even `enum Token`. But we then have to stub out a different `Context` definition and implementation for the tokenizer tests so they can still sort of collect tokens for their assertions.

I can start up an MR with those changes so we can have a look, but I don't think it should hold this simple performance win.

PR #128, opened 2025-01-11
Do not UTF8-encode individual characters when processing UTF8-encoded CDATA.

adamreichold:

PR #133, opened 2025-01-13
Optimize CDATA normalization using memchr and blockwise copying 

adamreichold:
`memchr` should be a well-known quantity as a dependency and provides ample optimization opportunities throughout the parser. For example, adjusting the normalization of line endings to use yields the following improvement for processing 1 MB of random CDATA:

```
before: test cdata_roxmltree                        ... bench:   6,039,358 ns/iter (+/- 41,232)
after:  test cdata_roxmltree                        ... bench:   3,143,947 ns/iter (+/- 168,323)
```

This would also supersede #128


adamreichold:
> one could implement the same chunking strategy using std's str::find (which however did not reliably apply SIMD the last time I tried this)

Gave this a try and it is still a significant improvement if not on par with `memchr`, c.f.

```
before:    test cdata_roxmltree                        ... bench:   6,039,358 ns/iter (+/- 41,232)
memchr:    test cdata_roxmltree                        ... bench:   3,143,947 ns/iter (+/- 168,323)
str::find: test cdata_roxmltree                        ... bench:   4,323,370 ns/iter (+/- 43,884)
```

RazrFalcon:
I don't mind `memchr` specifically, but the existing code is pretty slow on its own. We use `chars` + `enumerate` + `CharToBytes` - that's pretty slow. We might consider optimizing it first.
Also, CDATA it's an edge case, as far as I know. So unless we use `memchr` everywhere - it's not much of a benefit. 
Also, aren't `memchr` is used by Rust's std already?

Also, I'm not sure your code is equivalent. The existing one operates on Unicode codepoints, not bytes. We should make sure we're not breaking non-Latin XMLs.
PS: I don't remember why are going thought all those hoops with `TextBuffer` and `CharToBytes` to begin with. But I guess it was important at some point.

adamreichold:
So I added

```xml
<svg>
    <style><![CDATA[some
🐉
👃
]]></style>
</svg>
```

as a test case which requires line endings to be normalized and it behaves exactly the same on master as here.

RazrFalcon:
Got it. Agree. As long as we have a couple more tests and `lxml` produces the same output - we should be fine.
It was a long time since I last worked on this project.

For having nicer history I would suggest having two commits: 

1. Remove `TextBuffer` from `process_cdata`, aka simplify it + tests
2. Use `memchr` in `process_cdata`, aka opimization

adamreichold:
Will make it a two-step process, first without `TextBuffer` using bytes only and then using `str::find` (leaving out the `memchr` dependency to keep things simple).

I did have a look the existing CDATA tests and besides the one missing case of multi-byte characters before or after line breaks which I already added, I don't see any glaring omissions. Did you have some specific test case in mind?

adamreichold:
So added a simplified bytewise implementation with two additional test cases and then replaced it by the blockwise copying one without any changes in the output.

adamreichold:
Hhhmmm, so I had a look at the source of the standard library, c.f. <https://doc.rust-lang.org/stable/src/core/str/pattern.rs.html#439>, and while `memchr` is used for single-character patterns, it is still not specialized for single-byte patterns and the multi-character variants do not use `memchr` at all. So for example, we could not use `str::find` to meaningfully accelerate the normalization checks for text and attribute values. So, I think I'll go ahead and add another commit for `memchr` after all.

If had to go to a lonely island where survival would depend on writing fast parser and I could only take one crate, it would certainly be `memchr`...

RazrFalcon:
>Did you have some specific test case in mind?

A single multi-byte tests would be enough. As long as you think our tests cover all cases - it's fine.

>If I had to go to a lonely island where survival would depend on writing fast parser and I could only take one crate, it would certainly be `memchr`...

😄
I do agree. It's a nice bite-size crate.

adamreichold:
> A single multi-byte tests would be enough. As long as you think our tests cover all cases - it's fine.

Ok, I did add a new multi-byte test and another test with different combinations of `\r\n` and `\n` bunched up on the idea of catching off-by-one errors when copying the chunks. So I guess we should be fine for the time being.

PR #136, opened 2025-01-19
 Avoid quadratic runtime when merging text nodes

adamreichold:
With this change, the newly added `text` stress test/benchmark finishes in about a minute instead of I-did-not-wait-that-long. ~~However, it does also regress the real-world XML~~

```
 name              master ns/iter  branch ns/iter  diff ns/iter  diff %  speedup 
 huge_roxmltree    4,952,961       5,098,758            145,797   2.94%   x 0.97 
 large_roxmltree   2,316,864       2,477,716            160,852   6.94%   x 0.94 
 medium_roxmltree  577,389         573,742               -3,647  -0.63%   x 1.01 
 tiny_roxmltree    4,022           4,180                    158   3.93%   x 0.96 
```

~~which I am still looking into. I guess the common case of really just one borrowed text got slower.~~

The real-world XML files are basically flat

```
 name              master ns/iter  branch ns/iter  diff ns/iter  diff %  speedup 
 huge_roxmltree    4,940,395       4,848,394            -92,001  -1.86%   x 1.02 
 large_roxmltree   2,305,231       2,287,267            -17,964  -0.78%   x 1.01 
 medium_roxmltree  574,506         571,131               -3,375  -0.59%   x 1.01 
 tiny_roxmltree    3,998           4,002                      4   0.10%   x 1.00 
```

after I compressed the state space and made use of the fact that `StringStorage::new_owned` always creates a new allocation so that I can keep the original buffer around for if another text node need to be merged while still already leaving a fully formed text in the tree. (Note that these files never exercise the merging code path at all.)


RazrFalcon:
The code looks fine, but we have to reduce bench files size. It's way too big. Test files should ideally be less than 500KB.
And we should exclude `benches` from the Cargo package. No need to publish then on crates.io

adamreichold:
> The code looks fine, but we have to reduce bench files size. It's way too big. Test files should ideally be less than 500KB.

Adjusted the generator parameters and the three new files are well below 300 kB now.

> And we should exclude `benches` from the Cargo package. No need to publish then on crates.io

It is already exclude because it is a separate package, you can check by running `cargo package -v` which prints

```
warning: ignoring benchmark `xml` as `benches/xml.rs` is not included in the published package
```

among other things. (I did not exclude the tests as distribution often do like to be able to run tests for packaging.)

RazrFalcon:
Good.

PR #138, opened 2025-01-25
Speed-up attribute parsing by splitting tokenizing and verification

adamreichold:
This split allows us to use `memchr` to find delimiters and should also enable to better auto-vectorize the XML character verification.

This has a large benefit for the gigantic Jupiter SVG and small benefits and/or no impact for other benchmarks:

```
 name                  master ns/iter  memchr ns/iter  diff ns/iter   diff %  speedup 
 attributes_roxmltree  5,861,856       5,486,988           -374,868   -6.40%   x 1.07 
 cdata_roxmltree       314,965         315,581                  616    0.20%   x 1.00 
 gigantic_roxmltree    3,212,607       1,939,808         -1,272,799  -39.62%   x 1.66 
 huge_roxmltree        5,466,859       5,246,946           -219,913   -4.02%   x 1.04 
 large_roxmltree       2,430,924       2,359,054            -71,870   -2.96%   x 1.03 
 medium_roxmltree      637,052         576,896              -60,156   -9.44%   x 1.10 
 text_roxmltree        6,341,852       6,033,136           -308,716   -4.87%   x 1.05 
 tiny_roxmltree        4,404           4,456                     52    1.18%   x 0.99 
```

I did try to rework `is_xml_str` to be more enable to auto-vectorization but failed to produce any meaningful improvements. (I fear the predicate `is_xml_char` is just too branchy even in the ASCII case.) 

One strange thing I noticed is that add `#[cold]` to `gen_text_pos` made everything measurably slower. I think it would make sense to investigate setting the error positions after the fact as [in my pet fork](https://github.com/adamreichold/robinson/blob/main/src/tokenizer.rs#L135) to completely remove these calls from the hot paths. As long as the stream position is correct when the error return happens, this should still lead to exact error positions.


RazrFalcon:
Looks good.

About `rewind`, the current pattern is to remember the start position and then use `Stream::gen_text_pos_from`. `Stream` is intentionally forward-only. Mainly for architecture purity.

adamreichold:
> About `rewind`, the current pattern is to remember the start position and then use `Stream::gen_text_pos_from`. `Stream` is intentionally forward-only. Mainly for architecture purity.

Will fix.

RazrFalcon:
Looks good. 👍 

I don't really like that we have to pass a `Stream` to `is_xml_str`, but I'm not sure if there a nice way to avoid it...

PR #139, opened 2025-01-26
RFC: Drop tokenizer::Token intermediate data structure and XmlEvent trait

adamreichold:
As discussed previously here, this simplifies the code base, especially the interaction between tokenizer and parser. However, it also appears to imply that we have to drop the tokenizer test which intimately rely on the intermediate token representation.

Further code clean-ups are possible, e.g. standardizing on methods on Context instead of free functions, but this change aims for a minimal change while we discuss this.


RazrFalcon:
Looks good, but we **must** keep the tests. Removing them is not an option. Not sure what we can do about it.

adamreichold:
> Not sure what we can do about it.

The only thing that comes to my mind is conditional compilation, i.e. have a `cfg` flag that yields a completely different definition and implementation of `Context` which still collects tokens. But to be honest, considering that the devirtualization has already reaped most of the available performance benefits, I am not sure if this is worth it. Especially since it negates the main benefit of this change: simplifying the code.

PR #140, opened 2025-02-05
WIP: Add byte offset to `TextPos`.

Jayonas:
The goal here is to provide a byte position/offset in errors in addition to a row/col position.  The byte position is helpful for callers that want to do more sophisticated error reporting, such as:
1) showing source annotations using libraries like [annotate_snippets](https://github.com/rust-lang/annotate-snippets-rs), which naturally expect annotation positions as byte offsets,
2) using a more complex row/col calculation algorithm, such as accounting for single glyphs that use multiple chars / code points, or doing tab expansion.

The changes are pretty straight forward, and I only ran into one problem that I need feedback to resolve: there's a test that requires `size_of::<Error> <= 64`, but the new data in `TextPos` increases it to 72.  For now I commented that test out just to make sure all the other tests run and pass.  Any thoughts about this?  Can we just allow `Error` to get bigger?  Can we remove and/or shrink something else to compensate?  Can we potentially do the byte to row/col calculation lazily (e.g. as a method on `Error`) instead of storing row/col data in the error (I'm not sure that the necessary original data string would still be available, plus that could be a pretty break-y change)?  Can we hide this behind a feature flag?  Or does this error size limit potentially sink this PR/idea?

I also left a TODO in the `Display` impl of `TextPos` because I'm not sure if/how you might want to incorporate the byte position into that.

Sort of unrelatedly -- I added a new `TextPos::start` constructor for the not uncommon case where an error has no meaningful position and so just indicates the start of the data.  Seemed like a nice refactor, but it's obviously optional for my goal and I'm happy to change and/or revert that as desired.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -56,19 +56,35 @@ const XMLNS: &str = "xmlns";
 #[allow(missing_docs)]
 #[derive(Clone, Copy, PartialEq, Eq, Hash, Debug)]
 pub struct TextPos {
+    pub byte: usize,
     pub row: u32,
     pub col: u32,
 }
 
 impl TextPos {
     /// Constructs a new `TextPos`.
-    pub fn new(row: u32, col: u32) -> TextPos {
-        TextPos { row, col }
+    pub fn new(byte: usize, row: u32, col: u32) -> TextPos {
+        TextPos { byte, row, col }
+    }
+
+    /// Constructs a new `TextPos` that indicates the start of the data.
+    ///
+    /// Useful for errors that don't have a meaningful position in the data.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use roxmltree::*;
+    /// assert_eq!(TextPos::start(), TextPos::new(0, 1, 1));
+    /// ```
+    pub fn start() -> TextPos {
+        Self::new(0, 1, 1)
     }
 }
 
 impl fmt::Display for TextPos {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        // TODO: incorporate `byte` into `Display`?
         write!(f, "{}:{}", self.row, self.col)
     }
 }
@@ -194,8 +210,8 @@ impl<'input> Document<'input> {
     /// <e/>"
     /// ).unwrap();
     ///
-    /// assert_eq!(doc.text_pos_at(10), TextPos::new(1, 11));
-    /// assert_eq!(doc.text_pos_at(9999), TextPos::new(2, 5));
+    /// assert_eq!(doc.text_pos_at(10), TextPos::new(10, 1, 11));
+    /// assert_eq!(doc.text_pos_at(9999), TextPos::new(21, 2, 5));
     /// ```
     #[inline]
     pub fn text_pos_at(&self, pos: usize) -> TextPos {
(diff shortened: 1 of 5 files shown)


adamreichold:
> The changes are pretty straight forward, and I only ran into one problem that I need feedback to resolve: there's a test that requires size_of::<Error> <= 64, but the new data in TextPos increases it to 72. For now I commented that test out just to make sure all the other tests run and pass. Any thoughts about this? Can we just allow Error to get bigger? Can we remove and/or shrink something else to compensate? Can we potentially do the byte to row/col calculation lazily (e.g. as a method on Error) instead of storing row/col data in the error (I'm not sure that the necessary original data string would still be available, plus that could be a pretty break-y change)? Can we hide this behind a feature flag? Or does this error size limit potentially sink this PR/idea?

Due to the pervasiveness of `Result<(), Error>` in the tokenizer and parser code, the size of that type is highly relevant for the overall performance. Especially since we pay for the error case even when the majority of processed documents are valid.

Personally, I think this should be resolved by more aggressively pessimizing the error case and thereby improve the success case by boxing the error. We do depend on `alloc` even in `no_std` context so changing all relevant types to `Result<T, Box<Error>>` should not reduce the usefulness of the library while actually improving performance in the common case and making the size of `Error` itself largely irrelevant.

RazrFalcon:
We should either return a byte offset for error or text position. Not both.
I suggest returning just a byte offset and have a nice method to convert to "text position".
But that's would be a breaking change.

adamreichold:
>  I suggest returning just a byte offset and have a nice method to convert to "text position".

The problem I see is that this requires access on the original text, i.e. that method would effectively end up on `Document`, not on `Error`.

RazrFalcon:
We can have just a free function, like `roxmltree::text_position(string: &str, offset: usize) -> TextPos`.
Error reporting would be as nice out of the box, but it's better than having both values in the `Error`.

adamreichold:
Having both values there does not cost us anything, if we box the error.

RazrFalcon:
Having Box is an unnecessary complexity.

adamreichold:
It is also a performance optimization as it ensures `size_of::<Result<(), Box<Error>>() == size_of::<usize>()` thereby improving performance throughout call chains in the parser and tokenizer. This is the same reason why `anyhow` works so hard to avoid trait objects (and thereby fat pointers) to ensure `size_of::<anyhow::Error>() == size_of::<usize>()` using unsafe and raw vtables.

Admittedly, there is trade-off between added complexity and performance gain, but in my experience, I mostly interact with errors for forwarding and logging when calling into `roxmltree` and there, the `Box` does not make a difference at all.

RazrFalcon:
Oh, you're suggesting to put the whole `Error` into a `Box`, not just `TextPos`. Sure, it might help a bit, but I doubt we would see any performance differences.

But I still against storing byte offset in `TextPos`. We should store error position either as bytes or `TextPos`.

RazrFalcon:
I think we should start with replacing `TextPos` with `usize`. And maybe after that box errors, as a separate PR.

As for the API, just a free function should be fine. No need for a `TextPos` constructor and stuff.
And yes, we should remove `Document::text_pos_at`, since it will only confuse people.

Effectively, the new error handling approach should look like:

```rust
if let Err(error) = roxmltree::Document::parse(src_str) {
    let text_pos = roxmltree::text_pos(src_str, error.pos());
    eprintln!("Error: {} at {}", error, text_pos); // assuming we have changed `Display` for `roxmltree::Error` as well.
}
```

Issue #123, opened 2024-10-19
Maintenance help


RazrFalcon:
@adamreichold Me again. Would you be interested in maintaining this repo?
As you can see, the project is effectively done, so unless some major issues would arise there is no need to do anything with it. I just want to have someone as a backup in case I would not be able to access it anymore.

adamreichold:
The answer is the same as for `memmap2`, I am stretched but if the alternative is to archive, I will try to keep the lights on as I depend on this crate. It should be better here, since it is somewhat central at my current `$DAYJOB`, so I should be able to direct work time towards its maintenance.

RazrFalcon:
Thanks! I doubt you would need to worry about it. 1 commit per year would be enough 😄

PR #147, opened 2025-09-22
Add support for resolving external entities

### LINKED ISSUE #146: Looks like the `<!ENTITY name SYSTEM "foobar.xml">` does not work
Hello,

I want to parse some XML files with DTD like:

```xml
<!DOCTYPE doc [
<!ENTITY dnDisk SYSTEM "hdfs-site-dn.xml">
]>
```

And I set the `allow_dtd` field in options to `true` but get error below:

```
UnknownEntityReference(\"dnDisk\", TextPos { row: 81, col: 3 })
```

I try to search and find what happen, and I find code here:

```rust
// ExternalID ::= 'SYSTEM' S SystemLiteral | 'PUBLIC' S PubidLiteral S SystemLiteral
fn parse_external_id(s: &mut Stream) -> Result<bool> {
    let v = if s.starts_with(b"SYSTEM") || s.starts_with(b"PUBLIC") {
        let start = s.pos();
        s.advance(6);
        let id = s.slice_back(start);

        s.consume_spaces()?;
        let quote = s.consume_quote()?;
        let _ = s.consume_bytes(|c| c != quote);
        s.consume_byte(quote)?;

        if id == "SYSTEM" {
            // Ok
        } else {
            s.consume_spaces()?;
            let quote = s.consume_quote()?;
            let _ = s.consume_bytes(|c| c != quote);
            s.consume_byte(quote)?;
        }

        true
    } else {
        false
    };

    Ok(v)
}
```

It looks that the external XML file path are ignored? It is right?

adamreichold:
This adds minimal support for resolving external entities by enabling the user to supply a resolver which can act on the public ID and URI defining the entity.

Fixes #146

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Added
+- `ParsingOptions::entity_resolver` can be used to resolve external entities referenced via public ID and URI.
+
 ### Changed
 - `Node::has_attribute`, `Node::attribute` and `Node::attribute_node` now match local names similar to how `Node::has_tag_name` works.
 
--- a/src/parse.rs
+++ b/src/parse.rs
@@ -3,6 +3,7 @@ use alloc::{vec, vec::Vec};
 use alloc::borrow::Cow;
 use core::ops::Range;
 use core::mem::take;
+use core::fmt;
 use memchr::{memchr, memchr2, memchr_iter};
 
 use crate::{
@@ -130,6 +131,9 @@ pub enum Error {
     /// An invalid ExternalID in the DTD.
     InvalidExternalID(TextPos),
 
+    /// The given entity resolved yielded an errror.
+    EntityResolver(TextPos, String),
+
     /// A comment cannot contain `--` or end with `-`.
     InvalidComment(TextPos),
 
@@ -177,6 +181,7 @@ impl Error {
             Error::InvalidChar2(_, _, pos) => pos,
             Error::InvalidString(_, pos) => pos,
             Error::InvalidExternalID(pos) => pos,
+            Error::EntityResolver(pos, _) => pos,
             Error::InvalidComment(pos) => pos,
             Error::InvalidCharacterData(pos) => pos,
             Error::UnknownToken(pos) => pos,
@@ -187,7 +192,7 @@ impl Error {
 
 impl core::fmt::Display for Error {
     fn fmt(&self, f: &mut core::fmt::Formatter) -> core::fmt::Result {
-        match *self {
+        match self {
             Error::InvalidXmlPrefixUri(pos) => {
                 write!(f, "'xml' namespace prefix mapped to wrong URI at {}", pos)
             }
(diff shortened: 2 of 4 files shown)


adamreichold:
@PeterlitsZo Could you verify that this works for you? Thanks!

PeterlitsZo:
I am glad to say that this works well for my HDFS configure file. Thanks a lot.

@RazrFalcon Hello, do you think if this PR can merged?

adamreichold:
> Hello, do you think if this PR can merged?

Note that he said in #123 that he might not be able to work on this crate at all. I do want to give him a week or so to chime in and veto this, but without input, we'll have to find a way forward ourselves.

PeterlitsZo:
OK. Sounds great.

adamreichold:
> fmt::Debug for ParsingOptions looks ugly with that nested if. Better to extract it.

Done.

Kimahriman:
Do you have an example of how this could be used with loading additional data from disk? The unit test uses a static str, and it's not clear how to create owned data in entity resolution and return a reference to it

adamreichold:
> Do you have an example of how this could be used with loading additional data from disk? The unit test uses a static str, and it's not clear how to create owned data in entity resolution and return a reference to it

One option would be to use an arena, e.g.

```rust
let bump = bumpalo::Bump::new();

let mut entity_resolver = |pub_id: Option<&str>, uri: &str| {
    let path = uri_to_path(pub_id, uri);

    let buf = std::fs::read_to_string(&path)
        .map_err(|err| format!("Failed to read from {}: {}", path.display(), err))?;

    Ok(Some(bump.alloc_str(&buf)))
};
```

However, not that `uri_to_path` is very difficult to implement securely. A better approach would be to load all allowed external entities before starting to parse and putting them in a `HashMap` or something from which `entity_resolver` can just borrow.

Kimahriman:
Trying to do a flavor of that still leads to lifetime issues

```rust
let mut entity_resolver = move |pub_id: Option<&str>, uri: &str| {
    let buf = uri.to_string();
    let alloc: &str = bump.alloc_str(&buf);
    Ok(Some(alloc))
};
```

```
error: lifetime may not live long enough
   --> rust/src/common/config.rs:214:13
    |
211 |         let mut entity_resolver = move |pub_id: Option<&str>, uri: &str| {
    |                                   --------------------------------------
    |                                   |                                    |
    |                                   |                                    return type of closure is std::result::Result<Option<&'2 str>, String>
    |                                   lifetime `'1` represents this closure's body
...
214 |             Ok(Some(alloc))
    |             ^^^^^^^^^^^^^^^ returning this value requires that `'1` must outlive `'2`
    |
    = note: closure implements `Fn`, so references to captured variables can't escape the closure
```

adamreichold:
Sorry, the `move` did not belong there.

Kimahriman:
Without the `move`, `bump` doesn't live long enough

adamreichold:
You definitely need to borrow from `bump` and need to store it as far outside as necessary so that it lives at least as long as your input. This definitely works, c.f. the discussions over at https://github.com/adamreichold/robinson/pull/4.

Kimahriman:
That conversation was because of https://github.com/Kimahriman/hdfs-native/pull/260 which is what I'm trying to figure out how to improve. I think the only reason it works currently is because it uses raw pointers to create `&'static str`, which seems like a really hacky workaround to roxmltree not wanting to own anything. Still trying to figure out how `alloc_str` can work properly with the right lifetimes

PR #148, opened 2025-09-24
Adjust attribute accessors to match purely on local names

### LINKED ISSUE #145: `has_attribute` does not work with local name, unlike `has_tag_name`, which does
`has_tag_name` works with a local name, but `has_attribute` does not, which is inconsistent and surprising:

```rust
let doc = roxmltree::Document::parse("<root xmlns:foo=\"http://example.com/foo\"><foo:bar foo:baz=\"qux\"/></root>").unwrap();
let node = doc.descendants().last().unwrap();
assert_eq!(node.has_tag_name("bar"), true);
assert_eq!(node.has_attribute("baz"), false);
// similarly:
assert_eq!(node.attribute("baz"), None);
```

adamreichold:
This makes them less precise but more convenient. The manual approach of iterating attributes still exists, so this does not make any access patterns impossible.

Fixes #145

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
+### Changed
+- `Node::has_attribute`, `Node::attribute` and `Node::attribute_node` now match local names similar to how `Node::has_tag_name` works.
 
 ## [0.20.0] - 2024-05-23
 ### Added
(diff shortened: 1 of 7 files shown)


adamreichold:
> Just to clarify, Node::has_tag_name can skip namespace check when ExpandedName has no namespace, but Node::has_attribute and friends always check namespaces, right?

Exactly.

> Is this a breaking change? Because a new version would be able to return an attribute where the old version would return None?

Yes, this is a breaking change because of the reason you give.

> I guess it's worth doing to unify the API. No way I would remember why I did this 7 years ago.

I think so too. The main downside is that it becomes harder to actually force resolution without a namespace as shown in the modified test case. But I think this is also the least common use case.

RazrFalcon:
Feel free to merge.

PR #150, opened 2026-05-23
Fix a typo on the docstring for `EntityResolver`

musicinmybrain:
`s/errror/error/`


src/parse.rs
```diff
@@ -131,7 +131,7 @@ pub enum Error {
     /// An invalid ExternalID in the DTD.
     InvalidExternalID(TextPos),
 
-    /// The given entity resolved yielded an errror.
+    /// The given entity resolved yielded an error.
```
adamreichold: The typo fix seems correct. But the message still seems unclear overall, so maybe

```suggestion
    /// Resolving the given entity yielded an error.
```

musicinmybrain: Yes, that’s much more clear to me. (To be honest, I didn’t adjust the overall wording of the message because *I* wasn’t sure what it was trying to say.)

PR #155, opened 2026-09-20
Parse element content iteratively

### LINKED ISSUE #154: Stack overflow parsing deeply nested elements
`Document::parse` recurses once per level of element nesting, so a small document with deep nesting overflows the stack and aborts the process. 

`parse_element` and `parse_content` call each other with no depth bound.

It would be nice if we can limit the depth with an ParsingOptions::LimitDepth for instance

jaytaph:
Document::parse recursed once per level of element nesting without bounds and can result in a stack overflow (which will abort the application).

This commit changes this into a single loop so very deep nestings can be done safely without overflow, and it adds a limit configuration option which can be used to cap the depth.

Fixes #154


src/parse.rs
```diff
@@ -866,6 +888,13 @@ fn process_element<'input>(
         }
     }
 
+    // `parent_prefixes.len()` is the depth of the element about to be created.
+    if !matches!(end_token, tokenizer::ElementEnd::Close(..))
+        && ctx.parent_prefixes.len() > ctx.opt.depth_limit as usize
```
adamreichold: Instead of inferring the current depth from `parent_prefixes` and having a `depth` local in `parse_content`, wouldn't it be simpler to just have a `depth` stored directly in the parser?

I am also unsure about the use case for the configurable limit after this change avoids the stack overflow hazard? Deeply nested documents pay linearly in size for their nesting (except for entity expansion against which defenses already exist), so this does not seem like something that provides leverage and a simple size check on the input (which is always necessary when input is untrusted) should suffice.

jaytaph: The limit is an optional setting that we use in our code to track too deeply nested documents, but with this fix in place, we probably do not need it anymore since it should not be possible to trigger a stack overflow anymore with deep nesting. This can also remove this check on `parent_prefixes.len()`.

adamreichold: Sounds like a simpler and also importantly backwards-compatible change.

jaytaph:
Added additional commit. Can rebase it at a final stage when approved.

src/tokenizer.rs
```diff
@@ -678,9 +683,16 @@ pub fn parse_content<'input>(
...
                     parse_close_element(s, events)?;
-                    break;
+                    if depth == 0 {
+                        break;
+                    }
+                    depth -= 1;
+                }
+                Ok(_) => {
+                    if parse_start_tag(s, events)? {
+                        depth += 1;
```
adamreichold: `usize` could be small on some `no_std` platforms. Better to fix this to `u32` and use checked arithmetic on the increment code path. (You can err-out using the existing `NodesLimitReached` err variant if you adjust its docs.)

adamreichold (approved):

adamreichold (approved):
LGTM. (I am holding off merging for a bit to give others a chance to chime in.)

RazrFalcon:
Looks good. Except I wonder how long `deeply_nested_elements` takes to execute. According to GH Actions not as long, so I guess it's fine.

adamreichold:
> Except I wonder how long `deeply_nested_elements` takes to execute. According to GH Actions not as long, so I guess it's fine.

Takes less than 200ms on my puny Ryzen 5 5600U...

